PR #5, opened 2018-02-07
updated random number functions

olliemcdonald:
I changed the random uniform number generator function to use the one provided in PhysiCell_utilities to be consistent with the SeedRandom() function. Otherwise cells with random lifetimes on a single thread use the BioFVM version without specifying a seed, resulting in the same trajectory each time.


MathCancer:
This was also on our roadmap. Thank you!

PR #8, opened 2018-03-04
Update user_guide.tex

luissv7:
corrected spelling of preliminary from prelimanary and pointed out a place (DID NOT CORRECT) that a link does not show up correctly on pdf.


MathCancer:
Thanks, I'll give it a look now!

PR #21, opened 2019-06-13
2 bugfixes for out-of-bounds cells, and code cleanup

AndrewEckel:
- Cells that move outside of the boundaries in the process of dividing are now marked as out of bounds, avoiding a potential program crash.
- The assign_position function now assigns the position before checking whether the position is valid.  This corrects the 'center cell' bug, in which a new cell created in the divide function remains at (0,0,0) because the intended position is out of bounds.
- Eliminated unnecessary uses of 'continue'

### DIFF
--- a/core/PhysiCell_cell.cpp
+++ b/core/PhysiCell_cell.cpp
@@ -462,14 +462,6 @@ void Cell::set_previous_velocity(double xV, double yV, double zV)
 
 bool Cell::assign_position(double x, double y, double z)
 {
-	if( !get_container()->underlying_mesh.is_position_valid(x,y,z) )
-	{	
-		is_out_of_domain = true; 
-		is_active = false; 
-		is_movable = false; 
-		
-		return false;
-	}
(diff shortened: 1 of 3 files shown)


MathCancer:
Hi, sorry for delays. I'll work these bugfixes into 1.6.1. Thank you!

MathCancer:
I'm having a hard time fully comparing your updates to standard_update_cell_velocity in PhysiCell_standard_models.cpp.

I agree on avoiding "continue" wherever possible, but I'm reluctant to change this function at the moment. I'm defering further review of that particular fix until the next versoin so I can keep this moving. Thank you for the bug fixes!!!

MathCancer:
Also, sorry for how slow it was to integrate this into a release. Fall Semester got busy with going up for tenure, so I had to slow down code releases. 

Thanks again for your contribution!

MathCancer:
Thanks again for the bug fixes and cleanup!

PR #23, opened 2019-06-30
Differentiation code, sample projects, and makefile "reset" command bugfix

erifreem:
In first commit:

- Differentiation feature implemented, with changes to Cell and Phenotype core files

- Two new sample projects added, utilizing the differentiation functionality

- Makefiles altered to allow for usage of new sample projects

In second commit:
-  User Guide updated to explain differentiation functionality

In third commit:
- bug that prevented Makefile "reset" command from resetting "PhysiCell_settings.xml" fixed in all Makefiles


MathCancer:
We only accept pull requests into the development branch.

PR #30, opened 2019-12-28
fix location of default microenv

rheiland:
`16.2 Default microenvironment` - `microenvironment` is not in `core/PhysiCell_phenotype.cpp`; it's in `BioFVM/BioFVM_microenvironment.cpp`.

Wasn't sure if you wanted PRs on all intermediate Latex files and .pdf, but thought a cleaner workflow might be to just PR edits to the .tex and then build the pdf just before dev --> master.


MathCancer:
Sorry, just saw this now. I made the fix manually on my copy, since we're about to the release point/

Thanks for catchiing this!

PR #36, opened 2020-02-18
Bug fix on Necrosis Types

furkankurtoglu:
There is another "if check" in function of "update_cell_and_death_parameters_O2_based". 
Added:
- If "necrosis_type" is deterministic, the multiplier is a very big number.


MathCancer:
Furkan, what is the state of this? (I can't do an auto update.) 

Is deterministic working now? (One way to force deterministic is to set a very high transition rate if the deterministic condition is met.)

MathCancer:
PR was stale, but I made a fix to the bug in 1.7.0.

Thanks!

PR #48, opened 2020-05-26
Interface to use custom Cell classes

vincent-noel:
This PR adds an way to use custom cell classes, which are extensions of the Cell class.

We can cast from cell to custom cell back and forth in our custom functions, but the object needs to be instantiated using the custom cell constructor. 
To do this, we add a new custom function in Cell_Functions, to perform the instantiation : 
` 	Cell* (*instantiate_cell)(); `

We also add a standard_instantiate_cell, defined as such : 
```
Cell* standard_instantiate_cell()
{ return new Cell; }
```
When the pointer is null, this standard_instantiate_cell() function is called.

### DIFF
--- a/core/PhysiCell_cell.cpp
+++ b/core/PhysiCell_cell.cpp
@@ -129,7 +129,7 @@ Cell_Definition::Cell_Definition()
 		// the default Custom_Cell_Data constructor should take care of this
 		
 	// set up the default functions 
-	functions.instantiate_cell = standard_instantiate_cell;
+	functions.instantiate_cell = NULL;
 	functions.volume_update_function = NULL; // standard_volume_update_function;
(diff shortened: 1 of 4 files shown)


MathCancer (commented):
This looks like a pretty major change in the way we document and use, with more complex syntax. 

We've gone from: 

create_cell() or create_cell( cell_def )

to 

create_cell( somethign.something.some_method )

and we seem to be adding to cell_functions again. Let's do some thinking before we make changes that require major rework for most projects and don't appear to be backwards compatible. 

Will not merge into 1.8.0, but we can keep discussing.

MathCancer:
This does not appear to be needed anymore, since we successfully rolled out the unified architecture in 1.9.0.

PR #49, opened 2020-05-26
Custom adhesion and repulsion functions

vincent-noel:
This PR adds two new custom functions, to compute adhesion and repulsion. 

This time I left the original function if the pointer is null, and didn't create a "standard" function. But I can change that if you think it's better.

### DIFF
--- a/core/PhysiCell_cell.cpp
+++ b/core/PhysiCell_cell.cpp
@@ -849,8 +849,12 @@ void Cell::add_potentials(Cell* other_agent)
 		temp_a *= temp_a; // (1-d/S)^2 
 		// temp_a *= phenotype.mechanics.cell_cell_adhesion_strength; // original 
 		
-		// August 2017 - back to the original if both have same coefficient 
-		double effective_adhesion = sqrt( phenotype.mechanics.cell_cell_adhesion_strength * other_agent->phenotype.mechanics.cell_cell_adhesion_strength ); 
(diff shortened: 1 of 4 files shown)


MathCancer:
Huh, let's catch up in via video first.

MathCancer:
preferred design: 

Make a custom mechanics function that uses your preferred repulsion, etc. 
Then for some cell definition CD, set: 

CD.functions.update_velocity = noels_new_mechanics_function. 

Don't modify the existing functions or add a new function pointer to Cell_Functions unless completely necessary.

MathCancer (changes requested):
preferred design:

Make a custom mechanics function that uses your preferred repulsion, etc.
Then for some cell definition CD, set:

CD.functions.update_velocity = noels_new_mechanics_function.

Don't modify the existing functions or add a new function pointer to Cell_Functions unless completely necessary.

MathCancer:
What's the status on this? 

No longer needed, now that we have 1.9.0 out? 

I'm closing this if I don't hear otherwise. Thanks!

MathCancer:
Where do we stand on this one @vincent-noel ? 

Still needed?

vincent-noel:
When I made this PR, I was thinking it would be nice to only customize adhesion or repulsion functions. But as you said, it's probably unnecessary complexity, and just using a custom update_velocity is simpler.
You can close it.

PR #52, opened 2020-06-05
xml checks for phenotype motility

rheiland:


MathCancer (commented):
Really like this!

I might already be doing this in newer versions, so I'm going to manually compare first. 

THanks!

PR #56, opened 2020-11-17
Making sure we initialize the current mechanics voxel

vincent-noel:
There is a memory bug in the initialization of the cell's current mechanics voxel.
When we create the cell, we set it's volume, and try to update max_cell_interactive_distance_in_voxel. But since we don't know it's position yet, the current mechanics voxel is -1. 

The solution here is to : 
- check if current mechanics voxel is >= 0 before updating max_cell_interactive_distance_in_voxel.
- update max_cell_interactive_distance_in_voxel when we assign a position to the cell

This bug is actually silent most of the time, but might trigger a segfault sometimes (for some reasons which are not clear).

### DIFF
--- a/core/PhysiCell_cell.cpp
+++ b/core/PhysiCell_cell.cpp
@@ -517,6 +517,17 @@ bool Cell::assign_position(double x, double y, double z)
 	update_voxel_index();
 	// update current_mechanics_voxel_index
 	current_mechanics_voxel_index= get_container()->underlying_mesh.nearest_voxel_index( position );
+	
+	// Since it is most likely our first position, we update the max_cell_interactive_distance_in_voxel
+	// which was not initialized at cell creation
+	if( get_container()->max_cell_interactive_distance_in_voxel[get_current_mechanics_voxel_index()] < 
+		phenotype.geometry.radius * phenotype.mechanics.relative_maximum_adhesion_distance )
+	{
+		// get_container()->max_cell_interactive_distance_in_voxel[get_current_mechanics_voxel_index()]= phenotype.geometry.radius*parameters.max_interaction_distance_factor;
+		get_container()->max_cell_interactive_distance_in_voxel[get_current_mechanics_voxel_index()] = phenotype.geometry.radius
+			* phenotype.mechanics.relative_maximum_adhesion_distance;
+	}
+
 	get_container()->register_agent(this);
 	
 	if( !get_container()->underlying_mesh.is_position_valid(x,y,z) )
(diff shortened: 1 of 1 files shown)


MathCancer (commented):
@rheiland  can you give this a cross-platform test / double-check on 1.8.0-preview (dev branch)?

Looks really good! THanks!

rheiland:
I've examined this PR for some time now and will say the following:
1) if @vincent-noel had a test case that (easily) demonstrated this PR was necessary, it would be good to include it in our unit_tests (perhaps a problem arose during cell division?)
2) I tested the PR cross-platform and had no compile or (noticeable) runtime problems
3) using the function call in the index, i.e., `[get_current_mechanics_voxel_index()]` seems unnecessary (and adds a tiny bit of overhead) since we already have `get_current_mechanics_voxel_index` (unless I'm missing something).

vincent-noel:
I have a test case on this branch, using the 'physiboss-cell-lines' project from the sample project : https://github.com/sysbio-curie/PhysiBoSSv2/tree/bug/initialize_current_voxel_mechanics

I was never completely sure it was a real bug, or something I was missing during the initialization, so please have a quick look at this project. As far as I remember, it would only trigger a segfault on windows, and not always (I guess depending on the memory state at the -1 location of the vector). But now with the new tests I made where I explicitly clean memory at the end of the simulation, it crashes when freeing cells memory (and I'm not sure why yet). 

Link to the GitHub actions runs showing the bug : https://github.com/sysbio-curie/PhysiBoSSv2/actions?query=branch%3Abug%2Finitialize_current_voxel_mechanics

Link to the GitHub actions on the feat/intracellular branch, the only difference being this fix : https://github.com/sysbio-curie/PhysiBoSSv2/actions?query=branch%3Afeat%2Fintracellular

Valgrind finds it too : 
  
==29407== Invalid read of size 8
==29407==    at 0x17BA20: PhysiCell::Cell::set_total_volume(double) (in /home/vincent/Work/code/PhysiBoSSa/PhysiBoSS_Cell_Lines)
==29407==    by 0x1825F0: PhysiCell::create_cell(PhysiCell::Cell_Definition&) (in /home/vincent/Work/code/PhysiBoSSa/PhysiBoSS_Cell_Lines)
==29407==    by 0x1A87B9: setup_tissue() (in /home/vincent/Work/code/PhysiBoSSa/PhysiBoSS_Cell_Lines)
==29407==    by 0x11B6C9: main (in /home/vincent/Work/code/PhysiBoSSa/PhysiBoSS_Cell_Lines)
==29407== 

I'll have a better look to explain the crash when freeing memory. 
Also, you're completely right about using the variable current_mechanics_voxel_index instead of get_current_mechanics_voxel_index().

Vincent

MathCancer:
OK, thanks all. 

I think we pass on this PR for the time being, then. 

FWIW, I run windows daily and see ~0 seg faults on models I have built and run.

MathCancer:
This does not appear to be needed anymore, since we successfully rolled out the unified architecture in 1.9.0.

PR #66, opened 2021-03-03
fix Quickstart; comment out unused BioFVM function using rand()

rheiland:
Fix incorrect number of sample projects. Comment out "randomize" function in BioFVM_vector.cpp which we discovered in the reproducibility project.


MathCancer:
incompatible with more recent PRs. closing.

PR #77, opened 2021-10-20
A fix for setting Dirichlet conditions on boundary with nonempty symmetric differences

drbergman:
When the symmetric difference of the sets of Dirichlet conditions for two substrates is nonempty (so there are boundary points that are a DC for one but not both substrates), then the value in the config file for the non-DC substrate at such a point is used. This is in spite of the user setting that boundary as disabled for that substrate. This happens because at the end of BioFVM_microenvironment.cpp, a loop over all substrates to set whether they have DCs _also_ copies this vector of booleans to each voxel https://github.com/MathCancer/PhysiCell/blob/d45d290dd6e956bcc4c6899ff193748aad54bab4/BioFVM/BioFVM_microenvironment.cpp#L1527-L1530 which calls https://github.com/MathCancer/PhysiCell/blob/d45d290dd6e956bcc4c6899ff193748aad54bab4/BioFVM/BioFVM_microenvironment.cpp#L231-L239

Then, apply_dirichlet_conditions uses this voxel-specific value to determine whether to apply Dirichlet conditions for the given substrate at a given voxel https://github.com/MathCancer/PhysiCell/blob/d45d290dd6e956bcc4c6899ff193748aad54bab4/BioFVM/BioFVM_microenvironment.cpp#L279-L290

It is only because it first checks is_Dirichlet (see above) that this doesn't end up applying DCs across all voxels, only those that have a DC for some substrate.

The issue this causes can be seen by creating the template project and adding a substrate2. Using Dirichlet_options, set one substrate to have xmin enabled and the other to have xmax enabled (and all others disabled). Then, both substrates have a DC at both xmin and xmax. To make matters worse, if the _value_ at xmax is set for the first substrate, that will be the DC there **despite that boundary being disabled**.

To fix this, a new function is added to the Microenvironment class that just sets `dirichlet_activation_vector` for the object `microenvironment`
https://github.com/drbergman/PhysiCell/blob/60b38d540c4dceb4814691ef1610c3bd3e5239c2/BioFVM/BioFVM_microenvironment.h#L254-L255
https://github.com/drbergman/PhysiCell/blob/60b38d540c4dceb4814691ef1610c3bd3e5239c2/BioFVM/BioFVM_microenvironment.cpp#L241-L246

These leave `dirichlet_activation_vectors` alone which are already set elsewhere in BioFVM_microenvironment.cpp in lines such as 
https://github.com/MathCancer/PhysiCell/blob/d45d290dd6e956bcc4c6899ff193748aad54bab4/BioFVM/BioFVM_microenvironment.cpp#L1368-L1386

This fix should resolve this issue and simultaneously make it possible for others to use `get_substrate_dirichlet_activation` to determine whether a given substrate has a DC at a given voxel https://github.com/MathCancer/PhysiCell/blob/d45d290dd6e956bcc4c6899ff193748aad54bab4/BioFVM/BioFVM_microenvironment.cpp#L261-L262


MathCancer:
already integrated. thank you so much!

drbergman:
It seems like this bug persists. Using the latest release (v1.10.4), `make template`, and the attached config file (note: Github wouldn't accept an .xml so I include it as a .txt) reveals that the DC for substrate1 on xmax is actually being set to 1 despite it supposedly being disabled (Line 137). Similar issue with substrate2 on xmin (Line 156). A contour plot for substrate1 is also attached. Note: only the left boundary has DCs enabled for this substrate, so we would expect the concentration to decay from left-to-right and have a no-flux condition on the right boundary. Instead, it has its concentration fixed at 1 on the right boundary.
Screen Shot 2022-07-25 at 11 17 43 AM
[PhysiCell_settings.txt](https://github.com/MathCancer/PhysiCell/files/9182525/PhysiCell_settings.txt)

drbergman:
Just taking a look back at this to see if there's been any move to reopen. If it's easier, I can just open a new pull request on this.

drbergman:
TL;DR: I noticed a potential harm this could cause to others using `add_dirichlet_node`. I'll look into how that can be fixed.

Implementing this change could affect those currently using `add_dirichlet_node` in their custom code. That function only sets whether a voxel is a DC node. It does not change the activation_vectors that are checked before applying a DC condition. So, if someone relied on `activation_vector` being copied into each of the `activation_vectors` and then used `add_dirichlet_node`, this would cause all substrates with a boundary DC to have this value be a DC there. After my proposed change, that node would separately need to have been marked as a DC for the desired substrate(s).

PR #78, opened 2021-11-05
LibRR related bug-fixes and enhancements

furkankurtoglu:
- Correct Copasi File
- Better Cleaner MakeFile,
- gitingore changes for neglecting libRR downloaded source files.


MathCancer:
I think this is alreayd addressed, but we can revisit on more modern code. thanks!

PR #79, opened 2021-11-05
cancer cell cycle rate bug fix

rheiland:


MathCancer:
already addressed in 1.10.0. thank you for your constant vigilence!

PR #91, opened 2022-05-11
retouch jpeg-command in Makefile

saikiRA1011:


saikiRA1011:
The ">>" used in the jpeg command is the cause. Once the conversion fails, it will not work properly until __H.txt, __W.txt, etc. are deleted.
Converting ">>" to ">" prevented this problem.

MathCancer:
Thank you!
This is a good change. 
We don't do pull requests directly into master--only development. 
But I'll manually track this change through all the makefiles. 
Thank you!

PR #121, opened 2022-07-26
Update PhysiCell_cell.cpp

drbergman:
Fixed random orientation vector selection by using the standardized UniformOnUnitSphere function. Before, the z-value was uniform on [-1,1] which is not the desired uniformity.


drbergman:
Woops! I thought the code there was incorrect, but it turns out I was wrong and this is the right way to uniformly distribute points on a sphere. Sorry about that!

PR #129, opened 2022-10-06
Feat/svg substrate

marcorusc:
Hi Paul, Marco's here!

Here is the code to visualize a chosen substrate directly in the SVG. 
We discussed it the last time when me and Vincent were visiting you at IU.

It is still not perfect but at least we give a chance to the user to plot or not the substrate. 
During the hackathon at BSC people were interested in visualizing the substrate directly from the SVG, this is why I am finally doing the pull request (also because Vincent pushed me for months in doing it...).

Let me know what you think!

Best, 

Marco

### DIFF
--- a/modules/PhysiCell_pathology.cpp
+++ b/modules/PhysiCell_pathology.cpp
@@ -501,7 +501,7 @@ void SVG_plot( std::string filename , Microenvironment& M, double z_slice , doub
 		}
 		else
 		{
-			if(PhysiCell_settings.max_concentration != 0 | PhysiCell_settings.min_concentration != 0){
+			if(PhysiCell_settings.limits_substrate_plot){
 			 max_conc = PhysiCell_settings.max_concentration;
 			 min_conc = PhysiCell_settings.min_concentration;
 			}
--- a/modules/PhysiCell_pathology.cpp
+++ b/modules/PhysiCell_pathology.cpp
@@ -429,11 +429,24 @@ void SVG_plot( std::string filename , Microenvironment& M, double z_slice , doub
 		exit(-1); 
 	} 
 	
-	Write_SVG_start( os, plot_width , plot_height + top_margin );
+	if(PhysiCell_settings.enable_substrate_plot == true){
 
-	// draw the background 
-	Write_SVG_rect( os , 0 , 0 , plot_width, plot_height + top_margin , 0.002 * plot_height , "white", "white" );
+		double legend_padding = 200.0; // I have to add a margin on the left to visualize the bar plot and the values
 
+		Write_SVG_start( os, plot_width + legend_padding, plot_height + top_margin );
+
+		// draw the background 
+		Write_SVG_rect( os , 0 , 0 , plot_width + legend_padding, plot_height + top_margin , 0.002 * plot_height , "white", "white" );
+
+	}
+	else{
+
+		Write_SVG_start( os, plot_width , plot_height + top_margin );
+
+		// draw the background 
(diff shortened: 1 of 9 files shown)


rheiland:
Hi Marco, First, thanks for taking the initiative to do this!! I played with it a bit and have a few comments.
* I don't see any harm in providing such functionality in the future; however, I'd strongly suggest creating an additional method instead of modifying the existing `SVG_plot`; maybe do function overloading with the additional parameter.
* we'd probably want to agree on the new `<SVG>` elements in the .xml
* a user would need to know ahead of time what reasonable min_conc/max_conc should be; kind of challenging
* anytime a continuous field is color-coded, it's nice to also provide a colorbar showing the colors and numeric range
* it would be nice to add another parameter for cell opacity, e.g., `opacity="0.5"` (or whatever value) to append to each`<circle ...` in the .svg in order to better see the colored substrate below
* hopefully the PhysiCell Studio will provide more flexibility when rendering both cells and substrates (choose a substrate and min/max for colorbar); however, I too am a fan of vector graphics which the .svg provides, not to mention they can be easily viewed in a browser.

vincent-noel:
Hi Randy,

Glad you like it :)

>     * I don't see any harm in providing such functionality in the future; however, I'd strongly suggest creating an additional method instead of modifying the existing `SVG_plot`; maybe do function overloading with the additional parameter.

We thought it would simplify the main.cpp call to keep it in the same method. By default it's accessible, and only turned on/off in the XML.  
> 
>     * we'd probably want to agree on the new `<SVG>` elements in the .xml
> 

Yes this is just a draft, any changes are welcome.

>     * a user would need to know ahead of time what reasonable min_conc/max_conc should be; kind of challenging
> 

Actually we should have explained this, but if you remove the min/max, it's dynamic. It force us to add one for loop to check the min/max in every voxel, but it's working. Maybe the XML should be defined differently, with a on/off flag.

>     * anytime a continuous field is color-coded, it's nice to also provide a colorbar showing the colors and numeric range
> 

Excellent idea, we completely forgot about that ! Marco is on it :)

>     * it would be nice to add another parameter for cell opacity, e.g., `opacity="0.5"` (or whatever value) to append to each`<circle ...` in the .svg in order to better see the colored substrate below
>

Very cool idea.
 
>     * hopefully the PhysiCell Studio will provide more flexibility when rendering both cells and substrates (choose a substrate and min/max for colorbar); however, I too am a fan of vector graphics which the .svg provides, not to mention they can be easily viewed in a browser.

I think PhysiCell Studio could also use the XML settings in the end, to decide which substrate and if chosen which min/max ?

marcorusc:
Hello guys! 

I have updated the pull request adding a nice legend at the side of the plot.

Just a short explanation on how to set this in the xml:
In the "save" section of the xml file, under "SVG" I added the following block:

		<SVG>
			<interval units="min">30</interval>
			<enable>true</enable>
                        <plot_substrate enabled="true">
                            <substrate>toxin</substrate>
                            <min_conc>0.0</min_conc>
                            <max_conc>0.1</max_conc>
                        </plot_substrate>
		</SVG>

if you do not add min and max conc, it will force in the code a for loop that searches for the minimum and mac concentration
of the substrate that the user wants to visualize in the SVG (in the example above is toxin, the user can select here what to visualize).

If course if you disable the plot substrate, it comes back to normal svg visualization.

Finally, I am currently using the exponential format for the legend, but of course you can change it with whatever you prefer.
I hope you will like it!
g17818

MathCancer:
you taught me something new!

PR #134, opened 2023-01-26
fixes calcification dynamics

### LINKED ISSUE #133: [BUG] Calcified fraction (possibly) not calculated correctly in `standard_volume_update_function`
In [`PhysiCell_standard_models.cpp`](https://github.com/MathCancer/PhysiCell/blob/master/core/PhysiCell_standard_models.cpp)'s `standard_volume_update_function` I believe the calcified fraction update is not correct. All dynamic volumes except it are updated following a forward Euler method, e.g.

`phenotype.volume.fluid += dt * phenotype.volume.fluid_change_rate * 
		( phenotype.volume.target_fluid_fraction * phenotype.volume.total - phenotype.volume.fluid );`

The calcified fraction update does not have the `+` in the `+=`. It reads

`phenotype.volume.calcified_fraction = dt * phenotype.volume.calcification_rate 
		* (1- phenotype.volume.calcified_fraction);`

This means the cell will never calcify completely, it's calcified fraction will fluctuate around `dt * phenotype.volume.calcification_rate`

JulianoGianlupi:
Hello @MathCancer @rheiland and other PhysiCell devs, 

This small patch Closes #133 and fixes the calcification dynamics

### DIFF
--- a/core/PhysiCell_standard_models.cpp
+++ b/core/PhysiCell_standard_models.cpp
@@ -551,7 +551,7 @@ void standard_volume_update_function( Cell* pCell, Phenotype& phenotype, double 
 	phenotype.volume.total = phenotype.volume.cytoplasmic + phenotype.volume.nuclear; 
 	
 	
-	phenotype.volume.fluid_fraction = phenotype.volume.fluid / 
+	phenotype.volume.fluid_fraction += phenotype.volume.fluid / 
 		( 1e-16 + phenotype.volume.total ); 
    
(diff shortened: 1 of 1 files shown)


MathCancer (changes requested):
looks liek change to line 554 is a mistake. 

I'll make sure the fix to 548 makes it into my dev copy. THanks again for catching this!

MathCancer:
I made the relevant change to line 548 in my copy, so we don't need this particular PR. But thank you so much for finding (and fixing!) the bug!!!!

PR #141, opened 2023-03-14
set cell_positions folder to config.

elmbeech:
as stated by Heber Rocha, the folder name should be generic.


MathCancer:
Thanks for spotting this! 

I fixed it manually in my copy. Thank you!!!

PR #154, opened 2023-04-21
Update BioFVM_microenvironment.cpp

drbergman:
Not sure if `dirichlet_activation_vector` will be necessary going forward, but I think we want to default to `false` values for it as well.

Standardized additions to `Dirichlet_all` to be `false`. Previously, Line 594 had `false` and the rest had `true`.


MathCancer:
Already incorporated inot Randy's work now. Thank you, @drbergman !!

PR #162, opened 2023-06-14
Added custom functions to plot the agent SVG and legend

vincent-noel:
For PhysiMeSS, we need new design for our fibers in the svg files (a rectangle). To achieve this, we need to modify both the plotting of the svg plot, and of its legend. 

This PR introduce custom functions for the plotting of the SVG agent and of the agent's legend. The existing functions are now inside standard functions, and are still used automatically. But users now have the choice to choose a different function to plot them, by modifying the function pointer inside the Cell_Functions data structure.

### DIFF
--- a/core/PhysiCell_phenotype.h
+++ b/core/PhysiCell_phenotype.h
@@ -501,6 +501,11 @@ class Cell_Functions
 	void (*molecular_model_function)(Cell* pCell, Phenotype& phenotype , double dt ); 
 */
 	
+	
+	void (*plot_agent_SVG)(std::ofstream& os, Cell* pCell, double z_slice, std::vector<std::string> (*cell_coloring_function)(Cell*), double X_lower, double Y_lower);
+	void (*plot_agent_legend)(std::ofstream& os, Cell_Definition* cell_def, double& cursor_x, double& cursor_y, std::vector<std::string> (*cell_coloring_function)(Cell*), double temp_cell_radius);
+
+	
 	Cell_Functions(); // done 
 };
 
--- a/core/PhysiCell_standard_models.cpp
+++ b/core/PhysiCell_standard_models.cpp
@@ -67,6 +67,7 @@
 
 #include "PhysiCell_standard_models.h" 
 #include "PhysiCell_cell.h" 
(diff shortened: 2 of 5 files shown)


MathCancer:
I really like what you have here! And I see the logic for it. 

But if at all possible, I don’t think plotting belongs in PhysiCell_phenotype.cpp  and Cell_Functions. I wonder: is it possible to move it closer to PhysiCell_pathology.cpp

I guess I can see where you want to make how a cell is plotted part of the cell, so that differnt cells can be plotted different ways. that’s really clever.

I can see in the future we may want ot define this at the “type” level (as part of teh Cell_Definition . I’m not super keen on mixing visualization with actual cell biology, but I can see why it’s done here. Can we thikn if there’s a workaround?

For example a std::map between a Cell_Definition*  and a SVG_plot_function ? (and by default, that map is our standard SVG plot cell?)

MathCancer:
I admit this is cumbersome. We coudl auto-create the map by altering the constructor for `Cell_Definition` 

but I fear the mess. Hmm

MathCancer:
Let's approve this for now, but let's please consider a future revision that separates plotting from visualization. Thanks, Vincent!

furkankurtoglu:
Hello @vincent-noel 

ode-energy-sample model crashes at `SVG_plot()` function at following line.
https://github.com/MathCancer/PhysiCell/blob/development/sample_projects_intracellular/ode/ode_energy/main.cpp#L146


`my_coloring_function` is doing colorization based on custom_data which you may find here.
https://github.com/MathCancer/PhysiCell/blob/development/sample_projects_intracellular/ode/ode_energy/custom_modules/custom.cpp#L270

My question how can I accommodate this new `SVG_plot()` in 'ode-energy-sample' model? 

Thank you!

vincent-noel:
Hi @furkankurtoglu. I can look into this in a few hours. Which branch should I look at ? Is it already merged in the development branch ?

vincent-noel:
I found the problem, and proposed this fix : #186

furkankurtoglu:
It works! Thank you very much, @vincent-noel!

PR #194, opened 2023-08-01
Fixed inhibiting behaviors and cleaning. Updated version to 2.2.2

marcorusc:
Hello Paul! 

In this pull request I am correcting a bug that is preventing the inhibition of a behavior to work in PhysiBoSS. 
Previously, when an output node was connected to a behavior with action == inhibition, this was actually not working because the formula was not correct. Moreover, we removed the "test" variable from the if sentences when checking for the action, which was causing confusion and was not useful.

In a future pull request, I am planning to make the Hill exponent (now set to 10), as a parameter for the user, to give the user the possibility to choose between a step-like Hill function or a smoother one.

Please if you have any concern or doubt do not hesitate to let me know!

### DIFF
--- a/addons/PhysiBoSS/src/maboss_intracellular.h
+++ b/addons/PhysiBoSS/src/maboss_intracellular.h
@@ -10,7 +10,7 @@
 #include "maboss_network.h"
 #include "utils.h"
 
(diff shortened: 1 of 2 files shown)


MathCancer:
Thanks, Marco!

PR #197, opened 2023-08-01
Using a unique function to save both MultiCellDS data and intracellular data

vincent-noel:
This comes as an effort to make the main.cpp of PhysiBoSS (and ultimately intracellular) models standard. 
As of now, the only difference is that after saving the MultiCellDS data, I'm also saving a CSV PhysiBoSS state file. And if other intracellular models wants to save state, they have to do the same. 

<del>The idea here is to have a wrapper function to include MultiCellDS and intracellular data save, that I called ```save_PhysiCell_Timepoint``` and stored in ```PhysiCell_various_outputs```. </del>

<del>There will have no impact on the rest of PhysiCell, since in this PR only PhysiBoSS Cell Lines example is using this wrapper. But ultimately, it would be nice if all projects would use it, so that we get a standard main.cpp file. </del>

<del>There is no hurry for this, I just wanted to code it to see what it would look like, and have your opinion about this. </del>

At first I though about writing a wrapper to include the intracellular save with the full save, but after discussing with @elmbeech, and thinking more about this, I just need to add it to the MultiCellDS data. Here is what it would look like : 

```
<cellular_information>
	<cell_populations>
		<cell_population type="individual">
			<custom>
				...
				<intracellular_data type="text" source="PhysiBoSS" data_version="2">
					<filename>output00000002_boolean_intracellular.csv</filename>
				</intracellular_data>
				...
			</custom>
		</cell_population>
	</cell_populations>
</cellular_information>

```

The name of the tag, the filename, are completely open to suggestions.  
The filename use to be states_XXXXXXXX.csv, so it's already a breaking change for PhysiCell Studio and pcdl. I guess for them we would have to look at the XML file to see if the new intracellular_data field is there, otherwise check if the old format is there. 

Let's talk about this !

### DIFF
--- a/addons/PhysiBoSS/src/maboss_intracellular.cpp
+++ b/addons/PhysiBoSS/src/maboss_intracellular.cpp
@@ -473,14 +473,14 @@ void MaBoSSIntracellular::display(std::ostream& os)
 	std::cout << std::endl;
 }
 
-void MaBoSSIntracellular::save(std::string filename, std::vector<PhysiCell::Cell*>& cells)
+void MaBoSSIntracellular::save(std::string filename)
 {
 	std::ofstream state_file( filename );
 	
 	state_file << "ID,state" << std::endl;
 
-	for( auto cell : cells )
-		if (cell->phenotype.intracellular != NULL)
+	for( auto cell : *PhysiCell::all_cells )
+		if (cell->phenotype.intracellular != NULL && cell->phenotype.intracellular->intracellular_type == "maboss")
 			state_file << cell->ID << "," << static_cast<MaBoSSIntracellular*>(cell->phenotype.intracellular)->get_state() << std::endl;
 	
 	state_file.close();
--- a/addons/PhysiBoSS/src/maboss_intracellular.h
+++ b/addons/PhysiBoSS/src/maboss_intracellular.h
@@ -120,7 +120,7 @@ class MaBoSSIntracellular : public PhysiCell::Intracellular {
 
 	void display(std::ostream& os);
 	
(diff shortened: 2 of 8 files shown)


MathCancer:
on hold for now.

vincent-noel:
Ok I have a brand new proposal ! I updated the initial text

MathCancer:
Hmm, so if I'm reading this correct, this adds a `<custom>` sub-field to an existing `cell_population` field? 

This should work, although we may want to find a way to add it to the standard (for more general intracellular data) in the future. Do you have some sort of convention taht the list of cells in your `custom` is in the same order as in the "parent"?

vincent-noel:
Yes, it adds a new subfield, after the existing simplified_data, before the existing neighbor_graph fields. After thinking about it, I will rename it boolean_intracellular_data, so it won't step on the shoes of other intracellular addons. 
As for the order of the list of cells, it follows the order of all_cells, and there is the cell ID as the first column of the CSV file.

MathCancer:
Thanks. IN the future, maybe we will extend / stnadardize with something like: 

`<intracellular_data type="PhysiBoSS" data_version="1.0">` 

`<intracellular_data type="SBML" data_version="2.4">`

etc. Then we could have more in-depth discussion on how to standardize. But for now, in custom is the right place to put it. That's effectively a beta test until we're convinced it's a good design. 

Great work, and great suggestion from @elmbeech ! 

Thanks!

MathCancer:
checks look good. let's approve this PR. thank you!

elmbeech:
Thank you! Honestly, all credit to @vincent-noel!
I just mentioned to Vincent what would be best from the physicell data load point of view ": ).

PR #204, opened 2023-08-29
Make SVG plots with substrates easier and more configurable

drbergman:
Two main changes being suggested here:
1. **Make SVG plots with substrates easier**: Add the `substrate_coloring_function` to the template `main.cpp` and pass it into the three calls to `SVG_plot` so that all projects starting from there can include the substrate plot in SVGs just by configuring via xml or studio. If it is not set, `SVG_plot` already cleverly ignores all this, even if not specified. In other words, it should not cause errors.
2. **Make SVG plots with substrates more configurable**: Allow users to specify the colormap used in plotting the substrate in SVGs. Options include all those included in studio as well as their reverse values. Marco Ruscone's original colormap is included as `original`. Default, however, is `YlOrRd`. The list of colormap names is then {"original","YlOrRd","viridis","turbo","jet","plasma"} as well as any of these with a "_r" appended to indicate it is reversed.

Minor change: `substrate_coloring_function` originally returned a vector of 4 strings, probably because it started from the `cell_coloring_function` template which uses four colors. However, only one of `substrate_coloring_function` were used, so that is now reflected in the code.

Most of the lines added are simply the specifications for turbo, jet, and plasma. These have higher resolution because it was when I was adding those that I realized I could easily get Matlab to print out what I needed to specify these colormaps to arbitrary precision. If you edit in VS Code like a 21st century human, you can fold these out of view. Dinosaurs and vim users, I hope your editor-of-choice can do that.

Suggestions are welcome!

If accepted, this will be followed up with corresponding updates to studio.


MathCancer:
@vincent-noel  and @marcorusc  Do you anticipate any unexpected effects from this PR? I know that Marco in particular worked a lot on backgroudn coloring functions.

drbergman:
Follow up: the implementation in studio is already done and passed preliminary testing (just me).

Also, I guess I felt like throwing shade at the end there...not sure what prompted that. Sorry to all involved!

vincent-noel:
It looks very good to me : more options, simplification of the substrate_coloring_function. 

Looking at the list of commits, maybe you need to rebase your branch ? seems like many commits are not related to this change (and there are modifications in the readme which probably shouldn't be there). 

I need to test it, to triple check the existing examples still work, but again, thanks @drbergman, this is nice.

drbergman:
I just earned the `rebase` level 1 badge (did my first rebase). Not sure if I used it correctly, but I'd be happy to do so here if that will help in this setting.

I can go ahead and make sure the readme is the dev branch readme. I probably started from the main branch...

drbergman:
And since I'll be making changes to this anyways, I'll include the fix in #216

vincent-noel:
Good ! If everything looks good, then you need to force push the rebased branch, and we should see the list of commits updated (and reduced). Let me know if something goes wrong, I can try to help.

drbergman:
the easier solution (for me) was to create a new branch and then create a new PR. seemed safer than trying to make this work and possibly breaking things. See my new PR #219

PR #205, opened 2023-09-04
added ability to initialize BioFVM substrates with initial conditions

drbergman:
BioFVM substrates can be read in from a .mat file. The .mat file is a single matrix where each row is a voxel. The first three entries in the row are the x-, y-, and z- coordinates, respectively. The remaining columns are the values for ALL substrates.

Checks are made that the number of voxels (rows) is correct, number of substrates (columns - 3) is correct, and that the voxel indices corresponding to the positions in the (x,y,z) columns are all unique.

If the user wants to enable these, then simply go to the microenvironment options in the config xml, enable them, and specify the path to the .mat file relative to the PhysiCell directory.

Errors are issued with reminders about how to correctly set up the .mat file.

### DIFF
--- a/BioFVM/BioFVM_matlab.cpp
+++ b/BioFVM/BioFVM_matlab.cpp
@@ -108,7 +108,8 @@ std::vector< std::vector<double> > read_matlab( std::string filename )
      type_data_format > 5 || // unknown format
      type_matrix_type != 0 ) // want full matrices, not sparse 
  {
-  std::cout << "Error reading file " << filename << ": I can't read this format yet!" << std::endl;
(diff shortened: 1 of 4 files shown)


drbergman:
I tried renaming my branch for my own book-keeping sanity. Turns out that closed the previous PR. This one is identical except for the name of my branch.

drbergman:
Now includes ability to read CSVs. Must be formatted similarly to .mat files: one row per voxel with columns [x,y,z,substrate_0,substrate_1,...].

Possible feature improvement: header row to identify substrate names so order doesn't matter. Could also be used to allow for only a subset of substrates to be specified.

Issue #206, opened 2023-12-30
Sporadic error when rules sharing cell type and behavior


drbergman:
Sometimes, when two rules have the same target cell type and the same target behavior, a segmentation fault occurs when building the dictionary of rules for that cell type. This behavior has been observed for rules version 2, but may also occur when using other versions. These remain untested.

The error does not always occur when two such rules exist. In fact, in the two instances this error has manifested in the community, both were resolved with the same approach: move the offending rule up in the rules list. That is, when the segmentation fault occurs, it is immediately after the problematic rule is displayed in the console and so the user can identify which it is. Simply move this rule up the list (the top of the rules list will do) and the problem appears to go away.

Please post here if you have also ran into this issue. **If the solution above does not work for you, please post here and reach out; we will look into it asap.**

rheiland:
Have spent a lot of time looking into this. Bottom line, we need to get some of Paul's time to discuss the current implementation of rules. 

One minimal test case that demonstrates this issue is:
* create just one cell type, ctypeA
* provide the following rules (not intended to be biologically meaningful)

ctypeA,volume,decreases,migration speed,100.,200.,4,0
ctypeA,pressure,increases,cycle entry,1,5,4,0
ctypeA,pressure,increases,migration speed,1,5,4,0

The key point is there are two rules with the same behavior, migration speed, with different signals. But another rule separates them.

drbergman:
See PR #207 for a potential fix.

rheiland:
One bug that caused this problem was assuming the address to a std::vector's element was fixed. However, a vector dynamically allocates memory - it's not fixed. You can write a simple test program to see this. Therefore, this line:
```
             Hypothesis_Rule* pHR = &(rules.back());
```
is error-prone. In 1.13.1, it's at  https://github.com/MathCancer/PhysiCell/blob/master/core/PhysiCell_rules.cpp#L781C3-L781C42

There are a few different solutions, one of which is Daniel's (above), i.e., using `new` to allocate memory, then changing all references to pointers. Another might be to use the `reserve` feature for vectors. Another would be to use std::list (but then for all related params of rules). A discussion of the design and future plans for how rules may be used would be good, e.g., do we foresee the use of dynamic (run-time) rules or will they always be defined one-time at startup?

drbergman:
Would there be a performance hit for allowing them to change throughout a simulation? Obviously changing them will take some time, but will even just allowing them to change cause a performance hit?

[...]

PR #207, opened 2024-01-08
Fix bug in reading in rules.

drbergman:
Just needed to initialize the pointers in a better way for them to be passed across functions and objects. I won't pretend to understand why this works, only that it does work.

### DIFF
--- a/core/PhysiCell_rules.cpp
+++ b/core/PhysiCell_rules.cpp
@@ -768,17 +768,17 @@ Hypothesis_Rule* Hypothesis_Ruleset::add_behavior( std::string behavior , double
 		// if not, add it 
 	if( search == rules_map.end() )
 	{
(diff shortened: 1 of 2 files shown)


drbergman:
This is responsive to Issue #206

drbergman:
And in case there was any concern, I did look at all members of `Hypothesis_Rule` that are set in the Default Constructor and confirmed that `*pHR` has those values. OK, technically, I didn't check for the `std::vector<bool>` or `std::vector<std::string>` ones since they don't have a built-in way to interface with `<<`.

drbergman:
I have checked that with the second commit "Fixing the fix..." that the rules do in fact work, they are displayed as expected in standard output, and that it works on rules files that break PhysiCell currently.

PR #209, opened 2024-02-09
make Makefile jpeg and git functions work on Windows, Apple, and Linux.

elmbeech:
I used the `UNAME_S` variable and wrote `ifeq` `else`  `endif`  statements to make `make jpeg` and `make gif` work on Windows, Apple, and Linux.


MathCancer:
Full requests are only accepted into the development branch. Never main/master. 

GIF and JPEG have worked across platforms with properly installed ImageMagick, so I don't quite understand the purpose of this PR. In any event, PRs are never accepted to master except from development during a release.

elmbeech:
i see my mistake. i will make a pull request to the development branch!

it is not true that gif anf jpeg has worked across platform!
please have a look at the linux setup guide page 4: 
https://github.com/physicell-training/ws2023/blob/main/setup/PhysiCell_ws2023_Linux_setup.pdf 
the problem is that all linux distributions still run image magick version 7 for deep rooting reasons.
but the magick command was only introduced in image magick version 8. 
the are many linux user in the PhysiCell community, not only windows, apple m, and apple intel users. linux is one of the major platforms. we should support them too.

PR #210, opened 2024-02-11
make Makefile jpeg and git functions work on Windows, Apple, and Linux.

elmbeech:
I used the UNAME_S variable and wrote ifeq else endif statements to make make jpeg and make gif work on Windows, Apple, and Linux.

The problem is that many (all debian and centos based) Linux distributions still run image magick version 6 for deep-rooted reasons.
But the magick command was only introduced in image magick version 7.
There are many Linux user in the PhysiCell community, not only Windows, Apple M, and Apple Intel users. Linux is one of the major platforms. We should support them too, out of the box,


elmbeech:
These are the solutions I figured out so far:

+ make an alias like this  - but this is dangerous in my opinion because it will rewrite the Makefile each time  make is run:
```
alias make="sed -i -e s/magick//g Makefile; make"
``` 

+ make a magick command that  just pipes the rest of the command through, which is a very elegant solution in my opinion.
at `~/.local/bin/` (or where ever you have you local commands)
generate a `magick` file with the following content:
```
$*
```
then: `chmod 775 magick`

PR #213, opened 2024-02-21
Fix bug related to maximum/minimum saturation values and uniqueness test for rules

heberlr:
In the previous implementation, the last rule determined the maximum or minimum response of the behavior. In the current update, the behavior now checks for the maximum and minimum response in all rules. In cases where the minimum or maximum values do not exist, the system defaults to using the base value.

### DIFF
--- a/core/PhysiCell_rules.cpp
+++ b/core/PhysiCell_rules.cpp
@@ -798,8 +798,8 @@ Hypothesis_Rule* Hypothesis_Ruleset::add_behavior( std::string behavior , double
 
 Hypothesis_Rule* Hypothesis_Ruleset::add_behavior( std::string behavior )
 { 
-	double min_behavior = 0.1; 
-	double max_behavior = 1.0; 
+	double min_behavior = 9e99; // Min behaviour high value
+	double max_behavior = -9e99; // Max behaviour low value
 	return Hypothesis_Ruleset::add_behavior( behavior, min_behavior, max_behavior );
 }
 
@@ -1127,8 +1127,9 @@ void set_behavior_min_value( std::string cell_type, std::string behavior, double
             << ", but no rule for this behavior found for this cell type." << std::endl; 
         exit(-1);         
     }
-
-	hypothesis_rulesets[pCD][behavior].min_value = min_value; 
+	
+	if ( min_value < hypothesis_rulesets[pCD][behavior].min_value )
+	{ hypothesis_rulesets[pCD][behavior].min_value = min_value; }
 
 	return;
 }
@@ -1160,7 +1161,8 @@ void set_behavior_max_value( std::string cell_type, std::string behavior, double
         exit(-1);         
     }
 
-	hypothesis_rulesets[pCD][behavior].max_value = max_value;
+	if ( max_value > hypothesis_rulesets[pCD][behavior].max_value )
+	{ hypothesis_rulesets[pCD][behavior].max_value = max_value; }
 	
 	return;
 }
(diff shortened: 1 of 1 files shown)


drbergman:
I think this is something we want to address. It never feels right to assign a parameter value that is used only when the stars align. So, this seems like a good step in that direction. It looks like changes were not made for v0? I can't say I'm well-versed in that version, so not sure if that's an issue. It looks like v0 will use the last signal to set the min/max behaviors as well as the base_value?

On a similar note, I think a more longterm solution is to move to an XML-based rules specification. I have a branch with this implemented: https://github.com/drbergman/PhysiCell/tree/feature-xml-rules Happy to make a PR on this to MathCancer:development if others agree.

heberlr:
Thanks for pointing that out. I've incorporated tests into the v0 rules as well. During testing, I experimented with the rules listed below within the template project, making order changes, and selectively including or removing rules. I also checked the behavior by printing the values of min, base, and max in the evaluate() method of the Hypothesis_rules class, using an old-school debugging style.

rules v0:
default,cycle entry,0.00333333333333333333333333,0.00333333333333333333333333,0.033,pO2,increases,21.5,4,0
default,cycle entry,0.00333333333333333333333333,0.00333333333333333333333333,0.022,estrogen,increases,0.5,3,0
default,cycle entry,0,0.00333333333333333333333333,0.00333333333333333333333333,pressure,decreases,0.25,3,0

rules v1:
default,pO2,increases,cycle entry,0.00333333333333333333333333,0.033,21.5,4,0
default,estrogen,increases,cycle entry,0.00333333333333333333333333,0.022,0.5,3,0
default,pressure,decreases,cycle entry,0.00333333333333333333333333,0,0.25,3,0

rules v2:
default,pO2,increases,cycle entry,0.033,21.5,4,0
default,estrogen,increases,cycle entry,0.022,0.5,3,0
default,pressure,decreases,cycle entry,0,0.25,3,0

print code:
```
std::cout << "Min: " << min_value << "  Base: " << base_value << "  Max: " << max_value << std::endl;
	for( int j=0; j < signal_values.size(); j++ ){
		std::cout << "\tSignal: " << signal_values[j] << "  Response: " << responses[j] << std::endl;
```
P.S.: I believe that having the option to include rules in the XML could be very interesting.

heberlr:
As discussed, the rule is defined as unique if the cell, behavior, signal, and response are equal. Previously, the test only included the signal; in this pull request, we have added the response test. I test it with these rules:

default,substrate,increases,cycle entry,0.003333,0.2,10,0
default,substrate,decreases,cycle entry,0,0.5,10,0

This figure show that the rules is working (xml attached).
snapshot

drbergman:
Can you confirm that these cells are using the Live Cell model?

heberlr:
Yes, they are using live cell model. I will send you the xml and IC by slack.

core/PhysiCell_rules.cpp
```diff
@@ -253,31 +253,30 @@ void Hypothesis_Rule::add_signal( std::string signal , double half_max , double
...
 	// if so, then just warn and exit.  
-	if( n > -1 )
+	if( n > -1 && responses[n] == bResponse)
 	{
-		std::cout << "Warning! Signal " << signal << " was already part of the rule. Ignoring input." << 
+		std::cout << "Error! Signal " << signal << " and Response " << response << " were already part of the rule." << 
 		std::endl; 
 
-		return; 
+		exit(-1);
```
drbergman, line 274: I like that we're now throwing an error on this. Don't want to quietly disregard faulty user input

PR #216, opened 2024-03-04
Added missing check for substrate plotting

vincent-noel:
The legend of the substrate was added everytime, even when the substrate plotting was disabled in the XML.
This PR fixes it.


MathCancer:
No longer needed due to @drbergman

PR #218, opened 2024-03-04
Script to provide pre-compiled binaries, built by GitHub actions, for linux/mac/windows

vincent-noel:
This is still very preliminary, but I thought I would send a PR to start discussing it. 

I tried to find a way for users to avoid compilation, and to simply download binaries for a (small) list of projects.  The biggest use case are the template models : That way people can use them and simply use PhysiCell Studio to build their model. 
The binaries are build by GitHub actions, and are stored as release assets for each release. Then for users, they just need to run a python script, which will basically do the equivalent of make <project>; make. Meaning that the binary is ready, the config files are there, and if they want to add custom modules and recompile, it works. This script could later be integrated into PhysiCell Studio.
 
A few remarks about each OS : 

- On linux, it's relatively easy. The Makefile has to be modified to have static libopenmp, but nothing too ugly. 

- On macos, I'm building universal binaries (amd64 and arm64 compatible). One annoying issue here is that by default, macos will block downloaded binaries, so you have to allow them in the privacy and security settings. Also, sometimes the build fails, like 1/4 of the time. Not sure why yet, but just relaunching the job fixes it. On some machines (well, actually just one I know of), the binary runs via bash, but not via PhysiCell studio. And compiling works. 

- On windows, there are three dlls that I couldn't include in the binary, but that can be put in the same folder as the binary and it works. So I added a script to download these dlls. Not sure exactly how it works for license, but we are not providing PhysiCell with them, we just provide a way for the user to download them when needed. I probably should add their license somewhere in this case. Please don't sue me.

I'm including an example here with PhysiBoSS cell lines sample project, but I can quickly add another example if needed. 

What do you think ?

### DIFF
--- a/.github/workflows/build_binaries.yml
+++ b/.github/workflows/build_binaries.yml
@@ -50,7 +50,7 @@ jobs:
       with:
         upload_url: ${{ github.event.release.upload_url }}
         asset_name: ${{ matrix.projects.short_name }}-linux.tar.gz
-        asset_path: ${{ github.workspace }}/${{ matrix.projects.short_name }}.tar.gz
+        asset_path: ${{ github.workspace }}/${{ matrix.projects.short_name }}-linux.tar.gz
         asset_content_type: application/gzip   
         
         
--- a/.github/workflows/build_binaries.yml
+++ b/.github/workflows/build_binaries.yml
@@ -1,0 +1,262 @@
+name: Build binaries
+
+on:
+  push:
+  pull_request:
+  release:
+    types: [created]
+    
+jobs:
+         
+  ubuntu:
+    strategy:
+      matrix:
+        projects: [
+          {short_name: "physiboss-tutorial", project: "physiboss-tutorial", name: "PhysiBoSS tutorial", binary: "project", extra_run: "make Compile_MaBoSS PHYSICELL_CPP=g++-11"},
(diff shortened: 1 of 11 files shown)


MathCancer:
Waiting for an update ... (on hold for now)

rheiland:
First, *thanks* for doing this! I skimmed what you'd done then copy/edited it to suite my own needs of building just the template project. I haven't figured out (yet) how to do the macos universal build (just doing the arm64 now) - I lost patience, but will tackle it again sometime. And I also skipped the bit about having it auto-download the artifact. I just click on the link where it's at and download manually. Btw, did I interpret your workflow correctly that it would only be run for a new release? Would seem to make sense.

vincent-noel:
Hi Paul !

Indeed, this runs at every new release, and the binary are made publically available as release assets.
However, I was thinking that this could become a standard workflow in a repo dedicated to a model (I'm thinking of PhysiCell-Models here), where a new release would correspond to an update of a specific model. Then by using the identifier of the model repo, you could directly download the latest binary (and config folder) of this model. We should talk more about this. 

I will update this branch later next week, so that everything works and you'll be able to test the universal build. 
As you probably saw, one "issue" is that it needs modifications of the makefile, so have a static target. but nothing too complex.

PR #219, opened 2024-03-05
Substrate heatmaps in SVGs, defined in config file

drbergman:
This extends the functionality of plotting substrate heatmaps on SVG output:
* Allows for all colormaps permitted in studio for visualizing substrate concentrations
* Allows for configuration of this heatmap in the XML configuration file
* updates all sample projects so that they can use this functionality just by enabling SVG substrate plots in the XML

To utilize this, open a new project with the herein-modified `template` project and skip to Step 3 below. To make use of this in an existing project, do the following:
1. Add the following line in `main.cpp`. Immediately after defining `cell_coloring_function` works.
```
std::string (*substrate_coloring_function)(double, double, double) = paint_by_density_percentage;
```
2. Replace calls of `SVG_plot` in `main.cpp` with the following. There are likely **3** calls to replace.
```
SVG_plot( filename , microenvironment, 0.0 , PhysiCell_globals.current_time, cell_coloring_function, substrate_coloring_function );
```
3. Replace the `<SVG>` node with the following block to the configuration file, e.g. `config/PhysiCell_settings.xml`. The `colormap` element can be any of `YlOrRd`, `viridis`, `turbo`, `jet`, `plasma`, `original`, or any of these with `_r` appended to reverse the colormap.
```
		<SVG>
			<interval units="min">60</interval>
			<enable>true</enable>
			<plot_substrate enabled="true" limits="false">
				<substrate>substrate</substrate>
				<min_conc>0</min_conc>
				<max_conc>1</max_conc>
				<colormap>viridis_r</colormap>
			</plot_substrate>
		</SVG>
```

### DIFF
--- a/modules/PhysiCell_SVG.cpp
+++ b/modules/PhysiCell_SVG.cpp
@@ -100,6 +100,17 @@ bool Write_SVG_text( std::ostream& os, const char* str , double position_x, doub
   return true; 
 }
 
+void Write_SVG_text(std::ostream& os, const char* str , double position_x, double position_y, double font_size , const char* color , const char* font, double rotation)
+{
+    double text_width = font_size * strlen(str) / 2.0;  // estimate the width of the text
+    double text_height = font_size / 2.0;  // estimate the height of the text
+
+    double center_x = position_x + text_width / 2.0;
+    double center_y = position_y + text_height / 2.0;
+
+    os << "<text x=\"" << position_x << "\" y=\"" << position_y << "\" font-size=\"" << font_size << "\" fill=\"" << color << "\" font-family=\"" << font << "\" transform=\"rotate(" << rotation << " " << center_x << " " << center_y << ")\">" << str << "</text>\n";
+}
+
 bool Write_SVG_circle( std::ostream& os, double center_x, double center_y, double radius, double stroke_size, 
                        std::string stroke_color , std::string fill_color )
 {
--- a/modules/PhysiCell_SVG.h
+++ b/modules/PhysiCell_SVG.h
@@ -70,6 +70,7 @@
 #include <cstdio>
 #include <cstdlib>
 #include <string>
+#include <cstring>
 
 #ifndef _PhysiCell_SVG_h_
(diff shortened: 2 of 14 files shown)


drbergman:
Including previous discussions from PR #204 here:

# Original PR message:
Two main changes being suggested here:

Make SVG plots with substrates easier: Add the substrate_coloring_function to the template main.cpp and pass it into the three calls to SVG_plot so that all projects starting from there can include the substrate plot in SVGs just by configuring via xml or studio. If it is not set, SVG_plot already cleverly ignores all this, even if not specified. In other words, it should not cause errors.
Make SVG plots with substrates more configurable: Allow users to specify the colormap used in plotting the substrate in SVGs. Options include all those included in studio as well as their reverse values. Marco Ruscone's original colormap is included as original. Default, however, is YlOrRd. The list of colormap names is then {"original","YlOrRd","viridis","turbo","jet","plasma"} as well as any of these with a "_r" appended to indicate it is reversed.
Minor change: substrate_coloring_function originally returned a vector of 4 strings, probably because it started from the cell_coloring_function template which uses four colors. However, only one of substrate_coloring_function were used, so that is now reflected in the code.

Most of the lines added are simply the specifications for turbo, jet, and plasma. These have higher resolution because it was when I was adding those that I realized I could easily get Matlab to print out what I needed to specify these colormaps to arbitrary precision.

# Update on studio implementation
Follow up: the implementation in studio is already done and passed preliminary testing (just me).

# Inclusion of other PR changes
Changes made in #216 included here as well.

MathCancer:
On hold for some testing by Vincent

drbergman:
Here is my testing with it:
## `YlOrRd`
image

## `YlOrRd_r`
image

## `viridis`
image

## `viridis_r`
image

## `turbo`
image

## `turbo_r`
image

## etc
Also allowed: `jet`, `plasma`, `original`, and all these with `_r` suffix.

MathCancer:
Beautiful work, and cover dev discussion on slack confirms. Approved!

PR #230, opened 2024-05-19
changes to Makefile-default to allow user project custom_modules with a folder structure.

elmbeech:
To be able to pack my current projects as users projects, I would appreciate it if custom modules could have a folder structure. This small change to the Makefile-default would enable that. 
Thank you! Elmar

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -464,14 +464,14 @@ save:
 	cp main.cpp ./user_projects/$(PROJ)
 	cp Makefile ./user_projects/$(PROJ)
 	cp VERSION.txt ./user_projects/$(PROJ)
-	cp ./config/* ./user_projects/$(PROJ)/config
+	cp -r ./config/* ./user_projects/$(PROJ)/config
 	cp -r ./custom_modules/* ./user_projects/$(PROJ)/custom_modules
 
 load: 
 	echo "Loading project from $(PROJ) ... "
 	cp ./user_projects/$(PROJ)/main.cpp .
 	cp ./user_projects/$(PROJ)/Makefile .
-	cp ./user_projects/$(PROJ)/config/* ./config/ 
+	cp -r ./user_projects/$(PROJ)/config/* ./config/
 	cp -r ./user_projects/$(PROJ)/custom_modules/* ./custom_modules/
 
 pack:
--- a/Makefile
+++ b/Makefile
@@ -465,14 +465,14 @@ save:
 	cp Makefile ./user_projects/$(PROJ)
 	cp VERSION.txt ./user_projects/$(PROJ)
 	cp ./config/* ./user_projects/$(PROJ)/config
-	cp ./custom_modules/* ./user_projects/$(PROJ)/custom_modules
+	cp -r ./custom_modules/* ./user_projects/$(PROJ)/custom_modules
 
 load: 
 	echo "Loading project from $(PROJ) ... "
 	cp ./user_projects/$(PROJ)/main.cpp .
 	cp ./user_projects/$(PROJ)/Makefile .
 	cp ./user_projects/$(PROJ)/config/* ./config/ 
-	cp ./user_projects/$(PROJ)/custom_modules/* ./custom_modules/ 
+	cp -r ./user_projects/$(PROJ)/custom_modules/* ./custom_modules/
 
 pack:
 	@echo " "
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -465,14 +465,14 @@ save:
 	cp Makefile ./user_projects/$(PROJ)
 	cp VERSION.txt ./user_projects/$(PROJ)
 	cp ./config/* ./user_projects/$(PROJ)/config
```
drbergman, line 467: Might as well do the same for the config folder, too, yeah?

elmbeech: I don't think the config folder needs a deeper folder structure. 
There are usually just files in the config folder (PhysiCell_settings.xml, cell_rules.csv, cells.cvs). I don't need changes there.

drbergman: Sure. But since we're here, let's also change the behavior for the config folder, yeah?

elmbeech: on one hand, I don't care, on the other and I don't see a reason why to change that.
do you have an example, in which case you would need a deeper folder structure in config? 
do you have a user project that has configurations that have to be stored in a folder structure?
the case with the custom_modules is different. i have two modules I wrote, and each of these modules is in a folder, and it would be very tedious and result in a chaotic codebase, if i had to pack these modules into files only, no folders allowed.

vincent-noel: There is a need: we always put our boolean models in a folder.

MathCancer:
Let's verify it works on Windows one more time, and then approve. great change!

Makefile
elmbeech: @MathCancer befor you approve, I think I will then add @vincent-noel request too?!

vincent-noel: Thanks Elmar !

elmbeech:
I made the change that @drbergman and @vincent-noel asked for and tested it.
The Makefile-default can now as well handle config with folder structure.

rheiland:
Putting Windows testing aside for now, I'm confused about the premise of this PR. How does doing the `cp -r` change only in Makefile-default and not in every sample project's Makefile really achieve the goal? Am I missing something fundamental about the workflow of building a project?

Unrelated, the `physimess-sample` Makefile has no `load` or `save` target @vincent-noel.

vincent-noel:
I think I understand your point @rheiland, that these details are more important as a part of a project, and not in the "project loaders" that are Makefile and Makefile-default ?

Also, I added the missing Makefile instructions to PhysiBoSS and PhysiMeSS, in #239 and #240.

elmbeech:
@rheiland and @vincent-noel yes, it does solve the problem!
Because when you do a `make reset`, the `Makefile-default` gets loaded. 
And then you can load your user project with a folder deep `custom_modules` or `config` structure.
And if you have any other (sample) project loaded, you will certainly do a `make reset` before you load your peroject with deep folder structure.

rheiland:
We're kind of talking past each other @elmbeech . For the workflow you describe above, yes, it works as you expect. I was referring to the workflow of: make reset, make <sample_project>, make. That 2nd step copies over the Makefile from the <sample_project> which will not have the recursive `cp`. Agreed?

elmbeech:
@rheiland , but it does not matter that the Makefile from this <sample project> not have this cp -r changes, because when you will do make clean reset; make  <deep_folder_structure_project>, so you will have the correct Makefile back.
or are you usually directly loading a new project without doing a make clean reset before?

rheiland:
Again, you're fixated on the `make load PROJ` workflow. I'm not. I'm saying for the simple workflow of:
```
make clean reset
make template  # or any other sample project
```
you will not have a Makefile with the recursive copies for the `load` and `save`. That's it. That's all I'm saying. End of comments from me.

elmbeech:
but in this case you don't need it.

vincent-noel:
If you build over it, it might. Depends what the user will be doing next, which I think is what @rheiland is pointing to. 
We need it everywhere imo.

drbergman:
Adding these flags to all the sample projects will be simpler and faster than hashing this out. It won't break anything and it will ensure uniformity across all projects shipped with PhysiCell. That sound good to all?

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

elmbeech:
well, I can update all the sample projects if this is wished.
sure! 
will do it later today. i have now to go to the DMV (or BMV, how it is called in Indiana).

elmbeech:
This pull request was replaced by pr #244 , which has all Makefiles found in the PhysiCell repo changed accordingly.

PR #231, opened 2024-05-20
enable the random_seed integer parameter to be set for random behavior too.

elmbeech:
Currently, negative parameters.ints("random_seed") will cause the Mersenne Twister to be seeded with seed zero. 
There is no way to "seed" for random behavior.
For that reason, I saw code which randomly choosing an integer in a Python script to set the random_seed in the config/PhysiCell_settings.xml.
If my understanding how random code generators work is correct, then this approach is biased!
This is because the Mersenne Twister has 2^19937 - 1 states. Int64 however, has only 2^64 different states. Briefly, with this approach, these people miss out on most of the possible starting states.
There has to be an easy way set for random behavior!

With this small change in the template project, negative random seeds will result in random behavior, while zero and positive seeds will truly seed the random number generator, as we know it from the past.
This will be the case for any future projects, based on this template project.
For zero and positive numbers, nothing changes. We will not break with the past. But we open up a simple way to make random "seeding" for the runs possible for anyone, even directly in the Studio GUI, without having to make code changes in C++ in the custom.cpp.

### DIFF
--- a/sample_projects/template/custom_modules/custom.cpp
+++ b/sample_projects/template/custom_modules/custom.cpp
@@ -69,9 +69,13 @@
 
 void create_cell_types( void )
 {
-	// set the random seed 
-	SeedRandom( parameters.ints("random_seed") );  
-	
+	// set the random seed
+	if ( parameters.ints("random_seed") < 0 ) {
+		SeedRandom(); // random
+	} else {
+		SeedRandom( parameters.ints("random_seed") ); // seeded
+	}
+
 	/* 
 	   Put any modifications to default cell definition here if you 
 	   want to have "inherited" by other cell types. 


drbergman:
I completely agree that we should make the possibility for a time-based random seed (so we fully explore all of the Mersenne Twister seeds) a built-in feature of the template project. I would advocate for also checking if `random_seed` is a user parameter and if not, defaulting to `SeedRandom()`. The template project should not _require_ any one specific user parameter, in my opinion.

elmbeech:
Well, in the big picture, I agree with your argumentation.  
However, in the current template project implementation, random_seed is a given user parameter, and it is a required user parameter to run the template project's custom.cpp file.
Of course, you can make more changes to the template project or write a new template project from scratch.
However, with my pull request, I try to do as little changes to the current code base as possible to activate my goals.
I think this is best practice to not break too much with known behavior.

elmbeech:
I start to wonder if seeding the random number generator should really be part of `create_cell_types`, of if we not should make a one function for that in custom.cpp. 
e.g.:
```c++
void random_seed(void) {                                                        
    // set the random seed                                                      
    if (parameters.ints("random_seed") < 0) {                                   
        SeedRandom();  // random                                                
    } else {                                                                    
        SeedRandom(parameters.ints("random_seed"));  // fix                     
    }                                                                           
    return;                                                                     
} 
```

This function could then be called from the`main.cpp`, right before we call create_cell_types like blow:
```c++
[...]
random_seed();
create_cell_types();
[...]
```
Note: this is just a side thought, that possibly could be solved at the same go. 
However, the main reason for this pull request is not affected by this side thought.

MathCancer:
This is a great idea, but we dont' want ot leave it to user parameter or some convention. 

We propose adding a new setup variable in an appropriate section of the XML config file, and we'll update initialization options to use that.

elmbeech:
@MathCancer the setup variable in the XML sounds good to me too!

I just had a look at the `SeedRandom` code in `core/PhysiCell_utilities.cpp`. 
Since the code uses the system_clock to randomly seed the random number generator, I am not really sure if my other concern, that the Mersenne Twister has 2^19937 - 1 states and Int64 has only 2^64 different states, is addresses.

Reading up on the subject, it might be that we additionally have to use `std::seed_seq` to address this problem.

Here are some links who discuss this problem and possible solutions:
+ ( https://stackoverflow.com/questions/8395037/how-to-properly-seed-a-mersenne-twister-rng )
+ https://codereview.stackexchange.com/questions/109260/seed-stdmt19937-from-stdrandom-device
+ https://stackoverflow.com/questions/24334012/best-way-to-seed-mt19937-64-for-monte-carlo-simulations

with whom should I work together to look into this? 
should I make this an additional issue?

elmbeech:
@MathCancer this is now quite another approach as I suggested, and I don't see through how to get this resolved.
Could someone guide me to get this pull request in? Or if someone know what to do and feels dedicated, could he overtake?
thank you!

heberlr:
Please take a look at PR #252

drbergman:
[like]  Daniel Bergman reacted to your message:
________________________________

elmbeech:
this pr is superseded by #252

PR #232, opened 2024-05-20
avoid the possibility of having multiple parameters or multiple densities with the same name.

elmbeech:
In the current PhysiCell implementation, it is possible to generate multiple parameters with the same name. It is also possible to multiple substrates (density) with the same name.

I changed the modules/PhysiCell_settings.cpp add_parameter functions so that they first check if a parameter with this name already exists, before pushing a new entry to the container. 

I changed the BioFVM/BioFVM_microenvironment.cpp add_density functions so, that they first check if a density with this name already exist, before pushing a new entry to the containers.

This change will avoid having more than one parameter or density with the same name.

### DIFF
--- a/BioFVM/BioFVM_microenvironment.cpp
+++ b/BioFVM/BioFVM_microenvironment.cpp
@@ -473,212 +473,237 @@ void Microenvironment::resize_densities( int new_size )
 
 void Microenvironment::add_density( void )
 {
-	// fix in PhysiCell preview November 2017 
-	// default_microenvironment_options.use_oxygen_as_first_field = false; 
-	
-	// update 1, 0 
-	zero.push_back( 0.0 ); 
-	one.push_back( 1.0 );
-	
-	// update units
-	density_names.push_back( "unnamed" ); 
-	density_units.push_back( "none" ); 
+	// fix in PhysiCell preview November 2017
+	// default_microenvironment_options.use_oxygen_as_first_field = false;
 
-	// update coefficients 
-	diffusion_coefficients.push_back( 0.0 ); 
-	decay_rates.push_back( 0.0 ); 
-	
-	// update sources and such 
-	for( unsigned int i=0; i < temporary_density_vectors1.size() ; i++ )
+        // check if density exist
+	if (find_density_index( "unnamed" ) < 0)
 	{
-		temporary_density_vectors1[i].push_back( 0.0 ); 
-		temporary_density_vectors2[i].push_back( 0.0 ); 
-	}
 
-	// resize the gradient data structures 
-	for( unsigned int k=0 ; k < mesh.voxels.size() ; k++ )
-	{
-		gradient_vectors[k].resize( number_of_densities() ); 
-		for( unsigned int i=0 ; i < number_of_densities() ; i++ )
+		// update 1, 0
+		zero.push_back( 0.0 );
+		one.push_back( 1.0 );
+
+		// update units
+		density_names.push_back( "unnamed" );
+		density_units.push_back( "none" );
+
+		// update coefficients
+		diffusion_coefficients.push_back( 0.0 );
+		decay_rates.push_back( 0.0 );
+
+		// update sources and such
+		for( unsigned int i=0; i < temporary_density_vectors1.size() ; i++ )
 		{
-			(gradient_vectors[k])[i].resize( 3, 0.0 );
+			temporary_density_vectors1[i].push_back( 0.0 );
+			temporary_density_vectors2[i].push_back( 0.0 );
 		}
-	}
 
-	gradient_vector_computed.resize( mesh.voxels.size() , false ); 	
-	
-	one_half = one; 
-	one_half *= 0.5; 
-	
-	one_third = one; 
-	one_third /= 3.0; 
-	
-	dirichlet_value_vectors.assign( mesh.voxels.size(), one ); 
-	dirichlet_activation_vector.push_back( true ); 
-	dirichlet_activation_vectors.assign( mesh.voxels.size(), dirichlet_activation_vector ); 
-	
-	// Fixes in PhysiCell preview November 2017
-	default_microenvironment_options.Dirichlet_condition_vector.push_back( 1.0 ); //  = one; 
-	default_microenvironment_options.Dirichlet_activation_vector.push_back( false );
-	
-	default_microenvironment_options.initial_condition_vector.push_back( 1.0 ); 
+		// resize the gradient data structures
+		for( unsigned int k=0 ; k < mesh.voxels.size() ; k++ )
+		{
+			gradient_vectors[k].resize( number_of_densities() );
+			for( unsigned int i=0 ; i < number_of_densities() ; i++ )
+			{
+				(gradient_vectors[k])[i].resize( 3, 0.0 );
+			}
+		}
 
-	default_microenvironment_options.Dirichlet_all.push_back( true ); 
-//	default_microenvironment_options.Dirichlet_interior.push_back( true );
-	default_microenvironment_options.Dirichlet_xmin.push_back( false ); 
-	default_microenvironment_options.Dirichlet_xmax.push_back( false ); 
-	default_microenvironment_options.Dirichlet_ymin.push_back( false ); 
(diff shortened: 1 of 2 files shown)


drbergman:
This is a good improvement! Let's also take the opportunity to make this more maintainable and change the first two `add_density` implementations as follows:
```
void Microenvironment::add_density( void )
{
	// fix in PhysiCell preview November 2017 
	// default_microenvironment_options.use_oxygen_as_first_field = false; 
	return add_density( "unnamed" , "none" );
}

void Microenvironment::add_density( std::string name , std::string units )
{
	// fix in PhysiCell preview November 2017 
	// default_microenvironment_options.use_oxygen_as_first_field = false; 
	return add_density( name , units , 0.0 , 0.0 );
}
```

BioFVM/BioFVM_microenvironment.cpp
```diff
@@ -473,212 +473,237 @@ void Microenvironment::resize_densities( int new_size )
...
+		default_microenvironment_options.Dirichlet_xmin_values.push_back( 1.0 );
+		default_microenvironment_options.Dirichlet_xmax_values.push_back( 1.0 );
+		default_microenvironment_options.Dirichlet_ymin_values.push_back( 1.0 );
+		default_microenvironment_options.Dirichlet_ymax_values.push_back( 1.0 );
+		default_microenvironment_options.Dirichlet_zmin_values.push_back( 1.0 );
+		default_microenvironment_options.Dirichlet_zmax_values.push_back( 1.0 );
+        }
+        else
+        {
+		std::cout << "Warning: density " << name << " alredy exist!" << std::endl;
```
drbergman, line 703: 1. I think we ought to throw an error here. I know that's not strictly backwards compatible , but we should force the user to reckon with adding two identical substrates.
2. For readability, let's have the current `else` block be the first case (`if (find_density_index( name ) >= 0)`) so that we can throw an error (or `return` if that's what we choose) and then do away with the brackets afterwards

MathCancer:
I think this is a great idea, particularly in throwing an error in many cases. 

It seems like this is still under active diswcusssion / development, so let's not approve just yet.

elmbeech:
thank you @drbergman  for the suggestions. I will look into it.

elmbeech:
Hmm, throwing an error will cause a problem with the very project that made me aware of this problem.
In my project, during a run, I have to be able to reset parameters and densities to their initial values, if they already exist or generate a new parameter or density, if it does not already exist.
If we now through an Error, instead of a Warning, I gained nothing with this change.

heberlr:
@elmbeech The framework cannot be designed project specific, if we think about the future and troubleshooting simulations, throwing a error should be more appropriate, as this can avoid potential issues in duplication in the grammar keys.

elmbeech:
@heberlr in this case I do not agree.
it has to be possible to reset parameter and density structs to their initial values. 
and with an error instead of a warning this becomes impossible.

in any case, we talk here about a subject that was not a problem for the last 10 years at all.
now when I detected it and tried to fix it to solve my problem, you put another burden on me.

heberlr:
Let's try to list the pros and cons:
Pros:
- PhysiCell Studio already handles substrates with the same name. If you include substrates with the same name, the generated XML file has just one substrate.
- User parameters are currently used by customizable C++ code. Should we expect the C++ developer to check the warnings?

Cons:
- If you create two substrates with the same name in the .xml file and add rules associated with the substrate, the outcome won't be the expected one (throwing an error could help). 
Example:
Using the template project, add this rule:
`default,substrate,increases,apoptosis,0.1,0.5,10,0`
Try running two simulation with two different microenvironment setups:
1)
```
 <microenvironment_setup>
        <variable name="substrate" units="dimensionless" ID="0">
            <physical_parameter_set>
                <diffusion_coefficient units="micron^2/min">0.0</diffusion_coefficient>
                <decay_rate units="1/min">0.0</decay_rate>
            </physical_parameter_set>
            <initial_condition units="dimensionless">0</initial_condition>
            <Dirichlet_boundary_condition units="dimensionless" enabled="false">0</Dirichlet_boundary_condition>
            <Dirichlet_options>
                <boundary_value ID="xmin" enabled="True">0</boundary_value>
                <boundary_value ID="xmax" enabled="True">0</boundary_value>
                <boundary_value ID="ymin" enabled="True">0</boundary_value>
                <boundary_value ID="ymax" enabled="True">0</boundary_value>
                <boundary_value ID="zmin" enabled="False">0</boundary_value>
                <boundary_value ID="zmax" enabled="False">0</boundary_value>
            </Dirichlet_options>
        </variable>
        <variable name="substrate" units="dimensionless" ID="1">
            <physical_parameter_set>
                <diffusion_coefficient units="micron^2/min">1000.0</diffusion_coefficient>
                <decay_rate units="1/min">0.0</decay_rate>
            </physical_parameter_set>
            <initial_condition units="dimensionless">1</initial_condition>
            <Dirichlet_boundary_condition units="dimensionless" enabled="True">1</Dirichlet_boundary_condition>
            <Dirichlet_options>
                <boundary_value ID="xmin" enabled="True">1</boundary_value>
                <boundary_value ID="xmax" enabled="True">1</boundary_value>
                <boundary_value ID="ymin" enabled="True">1</boundary_value>
                <boundary_value ID="ymax" enabled="True">1</boundary_value>
                <boundary_value ID="zmin" enabled="False">1</boundary_value>
                <boundary_value ID="zmax" enabled="False">1</boundary_value>
            </Dirichlet_options>
        </variable>
        <options>
            <calculate_gradients>true</calculate_gradients>
            <track_internalized_substrates_in_each_agent>true</track_internalized_substrates_in_each_agent>
            <initial_condition type="matlab" enabled="false">
                <filename>./config/initial.mat</filename>
            </initial_condition>
            <dirichlet_nodes type="matlab" enabled="false">
                <filename>./config/dirichlet.mat</filename>
            </dirichlet_nodes>
        </options>
    </microenvironment_setup>
```
2)
```
        <variable name="substrate" units="dimensionless" ID="0">
            <physical_parameter_set>
                <diffusion_coefficient units="micron^2/min">1000.0</diffusion_coefficient>
                <decay_rate units="1/min">0.0</decay_rate>
            </physical_parameter_set>
            <initial_condition units="dimensionless">1</initial_condition>
            <Dirichlet_boundary_condition units="dimensionless" enabled="True">1</Dirichlet_boundary_condition>
            <Dirichlet_options>
                <boundary_value ID="xmin" enabled="True">1</boundary_value>
                <boundary_value ID="xmax" enabled="True">1</boundary_value>
                <boundary_value ID="ymin" enabled="True">1</boundary_value>
                <boundary_value ID="ymax" enabled="True">1</boundary_value>
                <boundary_value ID="zmin" enabled="False">1</boundary_value>
                <boundary_value ID="zmax" enabled="False">1</boundary_value>
            </Dirichlet_options>
        </variable>
        <variable name="substrate" units="dimensionless" ID="1">
            <physical_parameter_set>
                <diffusion_coefficient units="micron^2/min">0.0</diffusion_coefficient>
                <decay_rate units="1/min">0.0</decay_rate>
            </physical_parameter_set>
            <initial_condition units="dimensionless">0</initial_condition>
            <Dirichlet_boundary_condition units="dimensionless" enabled="false">0</Dirichlet_boundary_condition>
            <Dirichlet_options>
                <boundary_value ID="xmin" enabled="True">0</boundary_value>
                <boundary_value ID="xmax" enabled="True">0</boundary_value>
                <boundary_value ID="ymin" enabled="True">0</boundary_value>
                <boundary_value ID="ymax" enabled="True">0</boundary_value>
                <boundary_value ID="zmin" enabled="False">0</boundary_value>
                <boundary_value ID="zmax" enabled="False">0</boundary_value>
            </Dirichlet_options>
        </variable>
```
The dilemma is: Should we help the user build the model and prevent unexpected behavior by not allowing duplicate names, or should we expect them to look for the outcomes by checking the warning messages?

One argument in favor of your PR is that nowadays the likelihood of an user developing a model by editing the XML is very low.

elmbeech:
ok. i don't care.

elmbeech:
superseded by pull request #251. NF

drbergman:
I think you mean #250. but yes, this one can be closed

PR #233, opened 2024-05-20
Unnecessary warning message to "exit cycle phase 2 or 3" behaviors

heberlr:
Add a test for dead cells in the `set_single_behavior` function. This test avoids warning messages when using behaviors to exit cell cycle phases 2 or 3 in rules. Generally, this does not impact other function calls, as we do not need to change the cell cycle rates for dead cells. Note that even when using a rule not applicable to dead cells, the behavior values revert to their prior state (lines [372](https://github.com/MathCancer/PhysiCell/blob/84d033adba4a013f1fb3e7b7ece252ee254a1f26/core/PhysiCell_rules.cpp#L372) and [386](https://github.com/MathCancer/PhysiCell/blob/84d033adba4a013f1fb3e7b7ece252ee254a1f26/core/PhysiCell_rules.cpp#L386) in `PhysiCell_rules.cpp`)

### DIFF
--- a/core/PhysiCell_signal_behavior.cpp
+++ b/core/PhysiCell_signal_behavior.cpp
@@ -1260,7 +1260,7 @@ void set_single_behavior( Cell* pCell, int index , double parameter )
 
 	// cycle entry (exit from phase 0) and exit from up to 5 more phases 
 	static int first_cycle_index = find_behavior_index("exit from cycle phase 0" ); //  4*m; 
-	if( index >= first_cycle_index && index < first_cycle_index+6 )
+	if( index >= first_cycle_index && index < first_cycle_index+6 && !pCell->phenotype.death.dead )
 	{
 		int max_cycle_index = pCell->phenotype.cycle.model().phases.size(); 
 		if( index < first_cycle_index + max_cycle_index )


drbergman:
Clarifying question: you are calling `set_single_behavior` within your own custom code and not from a rule, correct?

heberlr:
This issue arises from the rules. To reproduce it: 
i) Populate the template project
ii) Create a rule using the behavior 'exit from cycle phase 2 or 3'. For example:
`default,substrate,increases,exit from cycle phase 2,0.04167,0.5,4,0`
If you want to reproduce it quickly, I recommend using a higher death rate ( in my test, I set the apoptosis rate to 5.31667e-04) and initializing with more cells (in my test, I set number_of_cells to 100).

As a result, you will get multiple warning message like that:
```
Warning: Attempted to set a cycle exit rate outside the bounds of the cell's cycle model
         Ignoring it, but you should fix this.
```

drbergman:
Oh! I get it now. A bit slow on the uptake. The signal can be toggled by `applies_to_dead` but not the behavior. Got it. I feel like this is a small issue with the rules: `exit from cycle phase n` should be rewritten to be specific to a certain cycle, not whatever cycle the cell finds itself in. Obvious convenience in writing it this way since you don't need to care about the cycle model being used. It's small because, as you noted, not many are going to have rules affect exit from death cycle phases.

vincent-noel:
We had this when building our PhysiBoSS controlled cell cycle. 
To me it's more of a warning that this rule shouldn't apply to dead cells, and that you forgot to turn this flag on. Agree with Daniel that ideally, there should be a distinct behavior for cell cycle and death cycle.

PR #238, opened 2024-05-31
allow user to change max num (spring) attachments in xml

drbergman:
allow for new element at end of `<mechanics>` element that can set the maximum number of attachments. Max num attachments is already a behavior so this is already hooked into the rules grammar.

Minimal working element:
```
<mechanics> 
	...
	<attachment_rate type="double" units="1/min">0.0</attachment_rate> 
	<detachment_rate type="double" units="1/min">0.0</detachment_rate> 
	<maximum_number_of_attachments>12</maximum_number_of_attachments>
</mechanics>
```


sample_projects/template/config/PhysiCell_settings.xml
```diff
@@ -266,6 +266,7 @@
                     <attachment_elastic_constant type="double" units="1/min">0.01</attachment_elastic_constant> 
                     <attachment_rate type="double" units="1/min">0.0</attachment_rate> 
                     <detachment_rate type="double" units="1/min">0.0</detachment_rate>
```
drbergman, line 268: I cannot explain why these few lines are unindented...

PR #241, opened 2024-06-04
Github action to build and run template project on all 3 OS

drbergman:
tests added for building and running the template project in all 3 OSs

NB: this is the first time I've made a github action. hopefully this is useful!

### DIFF
--- a/.github/workflows/test-macosx.yml
+++ b/.github/workflows/test-macosx.yml
@@ -67,3 +67,22 @@ jobs:
         ./project config/Fibre_Initialisation/mymodel_initialisation.xml
         ./project config/Cell_Fibre_Mechanics/mymodel_rotating.xml
         
+  build_template:
+
+    runs-on: macos-11
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Install dependencies
+      run : brew install gcc@11
+      
+    - name: Build template project
+      run: |
+        make template
+        make clean
+        make PHYSICELL_CPP=g++-11
+        
+    - name: Run template project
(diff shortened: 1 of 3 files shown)


elmbeech:
Nice! Thank you, Daniel!

I yesterday looked into the subject too.
GitHub actions are great, but I think this is not really scalable. 
We should combine GitHub action with an actual test framework.

I read up on test frameworks in my C++ book ( https://nostarch.com/cppcrashcourse ).
There are 3 test frameworks mentioned (although many more exist):
+ Catch2 ~ super lightweight, only a cpp header file ( https://github.com/catchorg/Catch2 ).
+ Google Test ~ heavyweight static library ( https://github.com/google/googletest ).
+ Boost Test ~ can by run by cpp header, as a static, or as a shard library ( https://github.com/boostorg/boost ).

Personally, I think Catch2 would be the way to go. 
It is popular, people say it is easy to write tests, it is lightweight, and I think will be scalable enough.

vincent-noel:
Agree with @elmbeech, the "tests" we have now are just compile and run a few projects, without any checks, and should ultimately be replaced by proper tests. 
Still, they are useful and we should already enjoy them. Maybe even add some basic check of the results. 

I have an idea how to code an action to make it very easy to add new projects, I'll give it a try and maybe submit a PR if I can make it work.

elmbeech:
This PR was merged into the [ template project code automatically generates output folder if it doesn't exist version2 #243 ]( https://github.com/MathCancer/PhysiCell/pull/243) pull request.

PR #243, opened 2024-06-04
template project code automatically generates output folder if it doesn't exist version2

elmbeech:
This PR should now work on all platforms and replaces [PR #235](https://github.com/MathCancer/PhysiCell/pull/235).
For more information, please read the discussion from 235.

In addition, this PR integrates pull request [Github action to build and run template project on all 3 OS #241](https://github.com/MathCancer/PhysiCell/pull/241)

Issue #220, opened 2024-03-08
Passing in files and folders as command line arguments


drbergman:
# Bottom line
Has anyone thought about passing in other files/folders at the command line rather than in the config file? I think it's going to be useful enough for me that I'll implement it in my own project, but just wondering if there's any advice out there to make it more readily merge-able into main.

# Background
I'm finding it would be helpful to be able to pass in paths to folders/files as command line arguments into a PhysiCell executable, similar to what's already done with the configuration file. Basically, I want to avoid changing the configuration file just to change the files/folders it encodes (output folder, initial conditions, rules). I can get it to work with minor edits to `main.cpp` for the output folder and configuration files, but it's a much bigger lift for the others requiring either
1. a rewrite of `create_cell_types` and `setup_tissue` which are essentially "core" code though live in `custom_modules`
2. a rewrite of `setup_cell_rules` and `load_cells_from_pugixml` which are `core` and `modules`, respectively

# Proposed solution
```
./project -c $(path_to_config) -o $(path_to_output) -i $(path_to_ic)
```

PR #246, opened 2024-06-05
Github action to build and run template project on all 3 OS and make output folder

drbergman:
This includes changes made in #241 (template project tests) and modifies the changes in #243 (make output folder). These are bundled together because of the intertwined history of those two PRs.

Unlike #243, this PR puts the output creation in `modules/PhysiCell_settings.cpp` to 
1. put it out of sight for most users since they won't need to do anything with it
2. make it accessible to others not using the template project `main.cpp`

I also went ahead and simplified the code around reading the path to the config file to reduce redundancy, improve readability, and enhance maintainability.

### DIFF
--- a/.github/workflows/test-macosx.yml
+++ b/.github/workflows/test-macosx.yml
@@ -67,3 +67,22 @@ jobs:
         ./project config/Fibre_Initialisation/mymodel_initialisation.xml
         ./project config/Cell_Fibre_Mechanics/mymodel_rotating.xml
         
+  build_template:
+
+    runs-on: macos-11
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Install dependencies
+      run : brew install gcc@11
+      
+    - name: Build template project
+      run: |
+        make template
+        make clean
+        make PHYSICELL_CPP=g++-11
+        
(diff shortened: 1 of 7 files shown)


elmbeech:
sounds all good to me! I close PR #243.
thank you, Daniel!

drbergman:
Some updates that will lead me to close this PR and open a new one (will link to that once opened)
* remove the gh actions update to make room for @vincent-noel's matrix of builds
* `create_output_directory()` will now create the full path to the output folder (e.g. `folder = output/subdir1`)
* add an `ArgumentParser` class to more elegantly handle args; this also dovetails with another idea I had (see #220) to allow other files to be passed in as args rather than in the config file

elmbeech:
This is already closed, but is there now output folder structure? with subfolders???
How far do this changes go?
This might cause heavily conflict with the physicell data loader!

elmbeech:
This is already closed, but is there now output folder structure? with subfolders???
How far do this changes go?
This might cause heavily conflict with the physicell data loader!

As well, no longer having this information in the settings.xml.
I think this should be more discussed!
This might break a lot for downstream analysis!!!

PR #247, opened 2024-06-05
updated scripts to build, run, test all sample projects

rheiland:
* this bash script compiles all (non-intracellular) sample projects and creates user_projects of each, preparing for the next step of running (brief) simulations of each.


drbergman:
I'm trying this out to test the arg parser and noticing the template project produces a file called `template` (on non-Windows) and the mechano-sample produces a file called `mechano-sample`. Subsequent runs of the script (after addressing issue, e.g.) then seem to confuse the make targets of `template` and `mechano-sample` for the newly created executables. Right now, I'm just appending an underscore to the executable name. bonus: that could be used to more easily delete at command line? Better is probably to prefix with an underscore so sorting puts them together.

rheiland:
Thanks very much! Pushed a fix. May go with a prefix or suffix underscore in the future, but for now staying with more traditional naming scheme. The header of the script lists all execs' names too.

rheiland:
I just pushed `beta/test_run_samples.py` to this PR which does what it says. It runs most sample projects after modifying each one's config file to only use 1 thread, brief max_time, and suffixed output dirs. The idea would be to run this on the previous release and the new [pre]release and compare results.

beta/test_run_samples.py
```diff
@@ -0,0 +1,64 @@
...
+#
+#   If you want to cleanup the created execs:
+# $ rm -rf template biorobots cancer_biorobots cancer_immune celltypes3 hetero pred_prey virus_mac worm interaction mechano rules physimess
+
+import subprocess
+import xml.etree.ElementTree as ET
+import os
+import time
+
+# skip over the 3D cancer_immune_sample for now
```
drbergman, line 34: add context so we know when to include that project in testing?

rheiland:
I think I'd like to keep them as separate scripts for now. I guess I should add comments in each to make it clear.

beta/test_run_samples.py
drbergman: since you said you were going to add comments to it, I would also appreciate clarity on this one :)

rheiland:
Just added a 3rd and final(?) script for testing: `beta/test_diffs_svg.py`. So, the order to run test scripts are:
* `beta/test_build_samples.sh`  (copy it to the root dir; see instructions in its header. Generates dirs in `/user_projects`)
* `python beta/test_run_samples.py`  # compiles/runs all projects in `/user_projects` generated from previous; edits models' xml so `max_time` is short, `omp_num_threads`=1, output `folder` is unique
* `python beta/test_diffs_svg.py  <dir1>  <dir2>`  # checks for bitwise equality of .svg files from two different repos/models, generated by previous script

MathCancer:
There appear to be no conflicts.

Issue #236, opened 2024-05-29
Fails to build with NVHPC 24.5


fspiga:
Hello, 

I am trying to build using NVIDIA C++ Compiler `nvc++` version 24.5
```
CC := nvc++
CFLAGS := -O3 -fomit-frame-pointer -mp -std=c++11
```

It complains with the following error(s):
```
nvc++ -O3 -fomit-frame-pointer -mp -std=c++11  -o heterogeneity BioFVM_vector.o BioFVM_mesh.o BioFVM_microenvironment.o BioFVM_solvers.o BioFVM_matlab.o BioFVM_utilities.o BioFVM_basic_agent.o BioFVM_MultiCellDS.o BioFVM_agent_container.o   pugixml.o PhysiCell_phenotype.o PhysiCell_cell_container.o PhysiCell_standard_models.o PhysiCell_cell.o PhysiCell_custom.o PhysiCell_utilities.o PhysiCell_constants.o PhysiCell_basic_signaling.o PhysiCell_signal_behavior.o PhysiCell_rules.o PhysiCell_SVG.o PhysiCell_pathology.o PhysiCell_MultiCellDS.o PhysiCell_various_outputs.o PhysiCell_pugixml.o PhysiCell_settings.o PhysiCell_geometry.o custom.o main.cpp
main.cpp:
/usr/bin/ld: PhysiCell_constants.o: in function `__sti___30___core_PhysiCell_constants_cpp_d2e95040':
nvc++Ormxb4w0R0PHy.ll:(.text+0xb4): undefined reference to `PhysiCell::PhysiCell_constants::advanced_Ki67_cycle_model'
/usr/bin/ld: nvc++Ormxb4w0R0PHy.ll:(.text+0xc4): undefined reference to `PhysiCell::PhysiCell_constants::advanced_Ki67_cycle_model'
/usr/bin/ld: nvc++Ormxb4w0R0PHy.ll:(.text+0x118): undefined reference to `PhysiCell::PhysiCell_constants::basic_Ki67_cycle_model'
/usr/bin/ld: nvc++Ormxb4w0R0PHy.ll:(.text+0x120): undefined reference to `PhysiCell::PhysiCell_constants::basic_Ki67_cycle_model'
/usr/bin/ld: nvc++Ormxb4w0R0PHy.ll:(.text+0x158): undefined reference to `PhysiCell::PhysiCell_constants::flow_cytometry_cycle_model'
[...]

PR #248, opened 2024-06-05
fix PhysiCell constants for more compilers

rheiland:
This may be premature, but would like to at least suggest it get merged in this release. This addresses compiler errors with `PhysiCell_constants{.h,.cpp}` for certain compilers (I had mentioned `clang++` when I need to do serious debugging, and issue https://github.com/MathCancer/PhysiCell/issues/236 mentioned the error on nvc++, as did @clasqui to @vincent-noel for PhysiBoSS). 

I've only tested with the "compile all the sample projects" script. Obviously, it would be good to test with actual simulations (EDIT: see next comment).

### DIFF
--- a/core/PhysiCell_constants.cpp
+++ b/core/PhysiCell_constants.cpp
@@ -76,12 +76,63 @@ double mechanics_dt = 0.1;
 double phenotype_dt = 6.0;
 double intracellular_dt = 0.01;
 
-std::unordered_map<std::string,int> cycle_model_codes = 
-{
-	{ "Ki67 (advanced)", PhysiCell_constants::advanced_Ki67_cycle_model}, 
-	{ "Ki67 (basic)" ,PhysiCell_constants::basic_Ki67_cycle_model},
-	{ "Flow cytometry model (basic)",PhysiCell_constants::flow_cytometry_cycle_model},
-	// { ,PhysiCell_constants::live_apoptotic_cycle_model}, // not implemented 
+
+// currently recognized cell cycle models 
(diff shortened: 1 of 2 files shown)


rheiland:
I used the 3 scripts discussed in https://github.com/MathCancer/PhysiCell/pull/247 to verify that the results generated using the modified versions of `core/PhysiCell_constants.{h,cpp}` in this PR's repo matched (bitwise)  the simulation results using the `core/PhysiCell_constants.{h,cpp}` in 1.13.1

Specifically, I ran this 3rd script (after running its 2 predecessor scripts to generate simulation results):
`M1P~/git/new-build-test$ python beta/test_diffs_svg.py ~/git/new-build-test/samples_test ~/git/fix-constants/samples_test`
and it confirmed that the (early) .svg results were bitwise reproducible for these sample models:
```
# note that we omit 'output_physimess' only because, currently, it fails the recursive copy of
# additional files needed in config/subdirs (until the Makefiles and "make load PROJ" is updated)

for out_dir in ['output_template','output_biorobots','output_cancer_biorobots','output_celltypes3',
'output_heterog','output_interaction','output_mechano','output_pred_prey','output_virus_mac',
'output_worm','output_rules','output_cancer_immune']:
```

rheiland:
I also confirmed that nvc++ compiles without any errors too (on Ubuntu). It's actually enlightening to see the warnings that it returns which hint at some code cleanup that could be done.

vincent-noel:
Works on my ubuntu too

PR #242, opened 2024-06-04
Simplification of the definition of projects to run in GitHub actions

vincent-noel:
I wrote a simpler version of the GitHub actions, using what's called a matrix: https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs

We define the list of projects to run like this:

```
projects: [
          {project: "virus-macrophage-sample", name: "Virus Macrophage", binary: "virus-sample"},
          {project: "physiboss-cell-lines-sample", name: "PhysiBoSS Cell Lines", binary: "PhysiBoSS_Cell_Lines"},
          {project: "physimess-sample", name: "PhysiMeSS Sample", binary: "project"},
          {project: "physimess-sample", name: "PhysiMeSS Sample", binary: "project", config: "config/Fibre_Initialisation/mymodel_initialisation.xml"},
          {project: "physimess-sample", name: "PhysiMeSS Sample", binary: "project", config: "config/Cell_Fibre_Mechanics/mymodel_rotating.xml"},
          {project: "template", name: "PhysiCell Template", binary: "project"}
]
``` 

Each element needs a project, which is the name used when running `make <project>`, a name, the filename of the binary file, and the optional path of a config file. 
This needs to be set up for all the different runners: macos-12 (x86_64), macos-14 (arm64), linux, windows. But the syntax is exactly the same.

PR #250, opened 2024-06-05
Throw error if duplicate substrate or parameter

drbergman:
Check that each added substrate has a unique name. Same for parameters (for a given type).

Also, clean up some code to improve maintainability/readability.

Suggestion for future improvement: rename some of the many instances of "parameters" in the code. For example, UserParameters has one instance called parameters. Each instance of Parameters has a member called parameters. Many other objects in PhysiCell have parameters members. This makes it hard to track which ones belong where.

### DIFF
--- a/BioFVM/BioFVM_microenvironment.cpp
+++ b/BioFVM/BioFVM_microenvironment.cpp
@@ -475,146 +475,24 @@ void Microenvironment::add_density( void )
 {
 	// fix in PhysiCell preview November 2017 
 	// default_microenvironment_options.use_oxygen_as_first_field = false; 
-	
-	// update 1, 0 
-	zero.push_back( 0.0 ); 
-	one.push_back( 1.0 );
-	
-	// update units
-	density_names.push_back( "unnamed" ); 
-	density_units.push_back( "none" ); 
-
-	// update coefficients 
-	diffusion_coefficients.push_back( 0.0 ); 
-	decay_rates.push_back( 0.0 ); 
-	
-	// update sources and such 
-	for( unsigned int i=0; i < temporary_density_vectors1.size() ; i++ )
-	{
-		temporary_density_vectors1[i].push_back( 0.0 ); 
-		temporary_density_vectors2[i].push_back( 0.0 ); 
-	}
-
-	// resize the gradient data structures 
-	for( unsigned int k=0 ; k < mesh.voxels.size() ; k++ )
-	{
-		gradient_vectors[k].resize( number_of_densities() ); 
-		for( unsigned int i=0 ; i < number_of_densities() ; i++ )
-		{
-			(gradient_vectors[k])[i].resize( 3, 0.0 );
-		}
-	}
-
-	gradient_vector_computed.resize( mesh.voxels.size() , false ); 	
-	
-	one_half = one; 
-	one_half *= 0.5; 
-	
-	one_third = one; 
-	one_third /= 3.0; 
-	
-	dirichlet_value_vectors.assign( mesh.voxels.size(), one ); 
-	dirichlet_activation_vector.push_back( true ); 
-	dirichlet_activation_vectors.assign( mesh.voxels.size(), dirichlet_activation_vector ); 
-	
-	// Fixes in PhysiCell preview November 2017
-	default_microenvironment_options.Dirichlet_condition_vector.push_back( 1.0 ); //  = one; 
-	default_microenvironment_options.Dirichlet_activation_vector.push_back( false );
-	
-	default_microenvironment_options.initial_condition_vector.push_back( 1.0 ); 
-
-	default_microenvironment_options.Dirichlet_all.push_back( true ); 
-//	default_microenvironment_options.Dirichlet_interior.push_back( true );
-	default_microenvironment_options.Dirichlet_xmin.push_back( false ); 
-	default_microenvironment_options.Dirichlet_xmax.push_back( false ); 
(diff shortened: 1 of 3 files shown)


elmbeech:
@drbergman, @heberlr, @MathCancer I am still voting for not thronging an Error, maybe not even a Warning.
The reason is, that:

+  `add_density` is solely called in the `PhysiCell_settings.cpp` `setup_microenvironment_from_XML` function.
+ `add_parameter` is solely called in the `PhysiCell_settinngs.cpp` `User_Parameters::read_from_pugixml` function.

This means, we only have to make sure, that if someone two times calls the high-level function `load_PhysiCell_config_file` (which e.g. is called in the main.cpp), that the variables are not twice generated.  Nothing else. 
Without error and warning, but checking for the condition, we handle this gracefully, without unnecessary breaking. 

This is the same as when you in Python write:
```
a = 1
a = 2
````
Python will not through an error and kill the runtime, it will simply overwrite the variable content.
Parameters and densities are high-level variables of PhysiCell and should, in my opinion, be handled accordingly.

**Comment by elmbeech** [comment]
If you, for whatever reason, insist that add_parameter and add_density through an error, then I request that at least the following: 

+ in the `PhysiCell_settings.cpp / setup_microenvironment_from_XML` function, the code should check if the density already exists, before `add_density` is called, and jump over it if the density already exists.

+ in `PhysiCell_settinngs.cpp / User_Parameters::read_from_pugixml` function, the code should check if the parameter already exists, before the `add_parameter` is called, and jump over it if the parameter already exists.

This would solve my problem too. But I think throwing no Error and no Waring is what we actually should do.

**Comment by drbergman** [comment]
These suggestions would walk back the main purpose of this PR: don't let users define a substrate (or parameter) and quietly not use it because it turned out to be a duplicate. We discussed the reasons for this change previously (#232). If you want to work towards a solution that fits within the constraints of this PR, I would suggest the following:
1. _if this is your own project_: just use the code you have now. it's ok if a project uses a variation on PhysiCell, in my opinion
2. _if this is a tool for others to use_: optional arguments that a user could overwrite to change the error -> warning within `add_density` and `add_parameter`. Then, the basic `main.cpp` can remain unchanged but users who want to downgrade the error to a warning can do so in `main.cpp`.

**Comment by elmbeech** [comment]
@drbergman, the main purpose of the original pull request (#232) (which actually was from me and not from you) was that parameters and densities variables can be reloaded without eating up memory, by every time generating a new variable with the same name. 
we discussed the reason, but the real issue why this pull request was generated in the first hand is not resolved and was not really address in the discussion because the people who were discussing it were not really understanding the problem. 

this pull request is here for a **tool for other to use**, and this is why it is critical that is properly resolved in the main code base.

**Comment by drbergman** [comment]
We discussed this over Zoom and in #232. I sided with you in #232. The rest of the community though agreed that this should result in an error. Heber had clear points on this in that thread. You asked for help amending that PR and that's why I made this one.

You have not acknowledged the issue we are trying to prevent with throwing an error, and that makes it difficult to continue this conversation. I have tried to ask why you need to loop through "episodes" in your `main.cpp` rather than running multiple instances and you declined to answer. So, to the extent I do not understand the problem, it is because you are withholding that information. I offered a possible solution above. Did you look into that? I'll offer another potential solution: move the `load_PhysiCell_config_file` call before your episodic loop. Since you have a single config file for all your episodes, you should not need to load the config file multiple times.

As for your python example, `add_density` clearly suggests adding a density to the list of current densities, i.e., not allowing for an overwrite. If the function were `add_or_update_density`, then that would properly indicate that the function could just update the density if it already existed.

I'm going to pause my own activity on this thread until others weigh in.

**Comment by elmbeech** [comment]
ok, i have now a version with add and update possibility and a flag that is by default set to add.
this should now make everyone happy.
have still fully to check that all works fine and think this is best tomorrow with a fresh mind. i should have a proper pull request tomorrow evening.

**Comment by elmbeech** [comment]
@drbergman I made a pull request with my changes to this branch on your PhysiCell fork.

**Comment by elmbeech** [comment]
this pull request is superseded by pull request #264 .

this version with add_density, add_parameter, update_density, and update_parameter functions should now make everyone happy. 
for more details, please read the comment in the #264  pr.

thank you, Elmar

**Comment by drbergman** [comment]
Great! I'll take a look once the checks are all passing on #264.

**Comment by elmbeech** [comment]
Thank you, @drbergman !

I looks like github has at this moment in general problem to run MacOSX jobs.
They are never picked up and executed.
When they  have fixed this, could you please re-run the jobs? I can't because I have no write permissions in the MathCancer PhysiCell repo.
+  https://docs.github.com/en/actions/managing-workflow-runs/re-running-workflows-and-jobs

Thank you, Elmar

**Comment by elmbeech** [comment]
@vincent-noel and @drbergman 
It looks that looks like macos 11 was removed for github actions. We have to switch to 12!

**Comment by elmbeech** [comment]
@vincent-noel 

This is the line that have to be changed in the yaml files:
```
runs-on: macos-11
```
Will try to figure out if there is something like `macos-latest`.

I will change it in my pull request. 
That would fix it for anyone, if accepted ; ).

elmbeech:
@drbergman , i just have a chat with @vincent-noel !
pull request #242 will fix the problem. 
as soon this pull request is accepted, the macosx check should pass too.

thank you.

elmbeech:
So, @drbergman  feel free to have a look at the pull request #264 already now ; ).

MathCancer:
We have philosophical qualms with this PR and are not convinced that this requires changes to PhysiCell core, or that it's a positive change. Deferring to future releases after additional code review.

PR #251, opened 2024-06-05
set default substrate color fn to give easy access to all projects

drbergman:
anybody can use the substrate coloring on svgs just by editing the config file, i.e. just by using studio. thanks @rheiland for the great suggestion!


elmbeech:
I guess this will supersede pull request #232; therefore I will close  #232.

PR #252, opened 2024-06-05
random seed within options

drbergman:
allow for node in `<options>` element that contains random seed info. see examples below. As discussed, the template project will continue to support the `random_seed` user parameter and will in fact use this to overwrite the random_seed in options. further remarks:
* `random_seed` intentionally used to match with user parameter to help draw attention for users
* a warning is issued if both are provided that someone other than me may one day see and convince them to keep only the random_seed in options

Random thought: we should write some of the console output to a file to focus attention on warnings that go flying by. also throttle updates on saves based on wall time passed.

# set by int (`SeedRandom( i )`)
```
<options>
...
    <random_seed>0</random_seed>
</options>
```
# set by system clock (`SeedRandom()`)
```
<options>
...
    <random_seed />
</options>
```
```
<options>
...
    <random_seed>system_clock</random_seed>
</options>
```
```
<options>
...
    <random_seed>random</random_seed>
</options>
```

### DIFF
--- a/core/PhysiCell_utilities.cpp
+++ b/core/PhysiCell_utilities.cpp
@@ -78,9 +78,21 @@ thread_local bool local_pnrg_setup_done = false;
 unsigned int physicell_random_seed = 0; 
 std::vector<unsigned int> physicell_random_seeds; 
 
-void SeedRandom( unsigned int input )
-{ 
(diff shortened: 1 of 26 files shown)


elmbeech:
this pull request supersedes #231

PR #244, opened 2024-06-05
changes to the cp command from make save and make load to allow user project config and custom_modules with a folder structure version2.

elmbeech:
This pull request now has the Makefile-default, the Makefile, all sample_projects Makefiles, all  
sample_projects_intracellular Makefiles, and all unit_test Makefiles changed accordingly.

Issue #222, opened 2024-03-29
Improvement: Makefile save command automatically adjusts PROJ := my_project 


elmbeech:
In each `Makefiel` is the following command

```bash
PROJ := my_project

save: 
        echo "Saving project as $(PROJ) ... "
        mkdir -p ./user_projects
        mkdir -p ./user_projects/$(PROJ)
        mkdir -p ./user_projects/$(PROJ)/custom_modules
        mkdir -p ./user_projects/$(PROJ)/config 
        cp main.cpp ./user_projects/$(PROJ)
        cp Makefile ./user_projects/$(PROJ)
        cp VERSION.txt ./user_projects/$(PROJ)
        cp ./config/* ./user_projects/$(PROJ)/config
        cp ./custom_modules/* ./user_projects/$(PROJ)/custom_modules
````

if the following `sed` line would be added,  then the  `PROJ := my_project` would automatically be set to what ever name was choosen for saving. 

e.g.: 
```
make save PROJ
```
would not change the line at all.

e.g.: 
```
make save PROJ=abc
```
would change the line to `PROJ := abc`

This is the `sed` line that would do the magic:
```bash
sed -i "0,/PROJ := /{s/PROJ := .*/PROJ := $(PROJ)/}" ./user_projects/$(PROJ)/Makefile
```

I added this line to the Makefiles of my PhysiCell user_projects, and it makes dealing with the projects much easier.

elmbeech:
issue #222 related to pull request #260 closed, based on the same reason.

Issue #212, opened 2024-02-21
make reset - avoid "rm *.cpp"?


rheiland:
Could we possibly *not* have `make reset` do a `rm *.cpp`? Just wondering what the danger is in having them? I've been bitten by this on multiple occasions. I make my own personal backup copy of a main.cpp, due to testing edits I've made, then it's removed on a make reset.

heberlr:
I'm not a huge fan of the "make reset" method. Maybe we can explore a setup where project folders containing the .cpp files are retained, and the main makefile calls the project-specific makefile within each folder, handling paths appropriately? I've encountered issues with "make reset," leading to the loss of a few hours of work. It could be more productive to collaborate and compile directly from the same .cpp file rather than relying on a copied version.

elmbeech:
i agree with Paul. i need `make restet` and `make clean` often, and they are for different tasks! 
however, when I do `make reset` i always have to do a `make clean before` because `make reset` will as well reset the particular `make clean` for the current project which might not be the same as in the default project.

elmbeech:
In my PhysiCell user_projects i added a line at the top of the `make reset` command to first call `make clean`.
I recommend this here as a default.

drbergman:
I would advocate for not changing the functionality of these common commands. I and others have come to rely on them and what they do. I don't see any of them as inherently broken. If we want, we can either extend any existing commands by introducing new `FLAGS` or add new commands, e.g. `make soft-reset`, `make save-reset`, `make clean-reset`. I don't think there's much downside in the Makefiles getting bloated. We just start users with the "base" commands and they can pick up new ones as they need.

elmbeech:
@drbergman i don't think it is a good idea to add more commands.
it is already hard enough to learn PhysiCell. 
more command that do nearly the same will confuse.
i believe in RISC we should boil down to best practice and a minimal instruction set.   

the point that i made is, that when you do a `make reset` without doing a `make clean` before, you might end up with orphan leftovers for the previous project!  
This means, currently, for a proper reset, you should never do a `make reset` without doing a `make clean` before. So why not call `make clean` inside the `make reset` function? This would eliminate the possibility of orphan leftovers.

drbergman:
See a comment at the (current) bottom of this PR #244 regarding whether `make clean` should be inside `make reset`. Link: https://github.com/MathCancer/PhysiCell/pull/244#issuecomment-2294103509.

[...]

PR #260, opened 2024-06-09
Makefile save function sets the PROJ variable in the Makefile to the actual name the project was saved with.

elmbeech:
This pull request is written on top of pull request #244 and addresses issue #222 .
What the pull request solves:

**The Makefile save function now sets the PROJ variable in the Makefile to the actual name the project was saved with.**

Before when e.g. a project was saved with
```
make save PROJ=abc
```
the save project had his PROJ variable still set to
```
PROJ := my_project
```
or whatever the PROJ variable was set before.
Now the saved project will have set the PROJ variable correctely to 
```
PROJ := abc
```
as it should be.

**This is a crucial change because it will prevent users from down the road overwriting accidentally the original project from which the newly saved project was derived from.**

@rheiland, because this changes use a sed command, cloud you, please check if this works on the Windows bash shell and on MacOSX zsh shell? That would be very helpful!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -490,8 +490,8 @@ save:
 	mkdir -p ./user_projects/$(PROJ)/custom_modules
 	mkdir -p ./user_projects/$(PROJ)/config 
 	cp main.cpp ./user_projects/$(PROJ)
+	awk -i inplace '{gsub("^PROJ :=.*$$", "PROJ := $(PROJ)", $$0); print $$0}' Makefile
 	cp Makefile ./user_projects/$(PROJ)
-	sed -i "0,/PROJ := /{s/PROJ := .*/PROJ := $(PROJ)/}" ./user_projects/$(PROJ)/Makefile
 	cp VERSION.txt ./user_projects/$(PROJ)
 	cp -r ./config/* ./user_projects/$(PROJ)/config
 	cp -r ./custom_modules/* ./user_projects/$(PROJ)/custom_modules
--- a/Makefile
+++ b/Makefile
@@ -394,9 +394,8 @@ reset:
 	touch ./custom_modules/empty.txt
 	touch ALL_CITATIONS.txt
 	rm ALL_CITATIONS.txt
-	mv ./config/PhysiCell_settings-backup.xml .
 	rm -fr ./config/*
-	mv PhysiCell_settings-backup.xml ./config/PhysiCell_settings.xml
+	cp ./sample_projects/PhysiCell_settings.xml-default ./config/PhysiCell_settings.xml
 
 clean:
 	rm -f *.o
--- a/Makefile
+++ b/Makefile
@@ -410,7 +410,7 @@ data-cleanup:
 # archival 
 
 checkpoint: 
-	zip -r $$(date +%b_%d_%Y_%H%M).zip Makefile *.cpp *.h config/*.xml custom_modules/* 
+	zip -r $$(date +%b_%d_%Y_%H%M).zip Makefile *.cpp *.h config/* custom_modules/* 
 	
 zip:
 	zip -r latest.zip Makefile* *.cpp *.h BioFVM/* config/* core/* custom_modules/* matlab/* modules/* sample_projects/* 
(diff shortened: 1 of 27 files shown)


rheiland:
The sed command does indeed seem to work on my Windows machine. But I'm still a bit confused. When I did `make save PROJ=foo`, it updates the Makefile in the foo user_project, but wouldn't you also want to update the root's Makefile so that one could then just do `make save` and have it update the foo user_project?

drbergman:
Looks like a very thorough set of changes and a nifty way to make it easier to use `make save`. I see you're also including `make clean` inside `make reset`. Did I miss the conclusion of #212?

elmbeech:
@rheiland, that is a good point. I can do the change to the Makefiles before I copy to the users_project folder.

@drbergman , there was no "conclusion" just expressed your opinion. 
if you do a reset, to work on a now project, it could simply cause errors if you have binaries from lying around from a project you worked before. or can you name me a PhysiCell project you work on, you still want to have the binaries around after a reset?  since i had to rewrite the whole function, I attempted to do a clean reset. and as i wrote in #212, it will not break your habits. you can do make clean, as much as you want, make clean reset as much as you want, and you can make rest as much as you want.

rheiland:
On my Mac, the native `sed` fails:
```
M1P~/dev/PhysiCell-patch_proj_name$ make save PROJ=foo
echo "Saving project as foo ... "
Saving project as foo ... 
mkdir -p ./user_projects
mkdir -p ./user_projects/foo
mkdir -p ./user_projects/foo/custom_modules
mkdir -p ./user_projects/foo/config 
cp main.cpp ./user_projects/foo
cp Makefile ./user_projects/foo
sed -i "0,/PROJ := /{s/PROJ := .*/PROJ := foo/}" ./user_projects/foo/Makefile
sed: 1: "./user_projects/foo/Mak ...": invalid command code .
make: *** [save] Error 1
```
I remember now that for a past project, I had to brew install the gnu sed to do what I wanted.

So this `sed`-related PR, as is, should not be merged.

elmbeech:
yes @rheiland, this is the exact same error we had last time.
I wonder if it is related to zsh. will track this down.
thank you for the report!

elmbeech:
well, we would have to figure out if the GNU sed or the BSD sed is installed.
this is getting `awkward` - so I will to try to find a way to achive the same with awk, hoping the GNU and the BSD awk will behave the same, and hoping that awk is installed on Windows.

elmbeech:
@rheiland i just pushed the changes. 
could you please give another try?

rheiland:
On my  Mac, native `awk` fails:
```
(base) M1P~/git/elmar_patch_proj_name$ make save PROJ=hetero
echo "Saving project as hetero ... "
Saving project as hetero ... 
mkdir -p ./user_projects
mkdir -p ./user_projects/hetero
mkdir -p ./user_projects/hetero/custom_modules
mkdir -p ./user_projects/hetero/config 
cp main.cpp ./user_projects/hetero
awk -i inplace '{gsub("^PROJ :=.*$", "PROJ := hetero", $0); print $0}' Makefile
awk: unknown option -i ignored

awk: can't open file {gsub("^PROJ :=.*$", "PROJ := hetero", $0); print $0}
 source line number 1
make: *** [save] Error 2
```
```
(base) M1P~/git/elmar_patch_proj_name$ which awk
/usr/bin/awk
(base) M1P~/git/elmar_patch_proj_name$ awk --version
awk version 20200816
```

elmbeech:
why the heck Berkeley Software Distribution?

elmbeech:
at least on Windows it seems to work. 
Will try to work on a workaround that works on all OSes.

MathCancer:
I don't understand the purpose of this PR and change. 

I think it's a *bad* idea to overwrite the Makefile. 

PROJ defaults to my project unless you use a command-line argument. The command line argument always works. What problem are we trying to solve here?

elmbeech:
ok. I think you understood what my intention was.

i always understood a 
+ `make save PROJ=abc` as `save as`
+ `make save` as `save`
obviously, this is not how you want to run things.

(look in now over the whole Makefile, i wonder why we actually have two variables `PROGRAM_NAME` and `PROJ`. Would it not simplify things, if this was one single variable?)

anyhow,  I will drop this pull request but keep the  #244 pull request alive, which solves the deep folder structure. 

thank you, Elmar

PR #261, opened 2024-06-19
Allow Turning Off Diffusion Solver for Substrates

heberlr:
This pull request introduces the ability to disable the diffusion solver for substrates. Users can now have multiple substrates without performing the computation of the BioFVM equation, which speeds up simulations that involve non-diffusible substrates, such as ECM, matrigel, or collagen.

This is backward compatible: The default behavior remains unchanged, running the diffusion_decay_solver for both LOD_2D and LOD_3D.

To turn off the diffusion solver, use the following variable in the XML configuration:
```
    <microenvironment_setup>
        <variable name="substrate" units="dimensionless" ID="0">
            ...
        </variable>
        <options>
            <solve_diffusion_decay>false</solve_diffusion_decay>
           ...
    </microenvironment_setup>
```


MathCancer:
suggest renaming / reparsing as `disable_diffusion_decay_solver`

PR #262, opened 2024-06-20
Exploring a log file to focus console output

drbergman:
**This PR needs feedback before merge**

I know I do not pay attention to the console output at startup mostly because it goes by so fast. I barely know what shows up there (Did you know it prints "rulesets" every time it finds a `rulesets` element?). Warnings are instantly lost and not received by the user. Rulesets not enabled? You'll never know until you check the results. The console scrollback buffer can also be quickly exceeded and lose some of that opening info.

So, I am looking to have a conversation around creating a log file that puts some/most of this info into a log file to allow the user to focus on more critical output. I started with this PR that aggressively swings the pendulum the other way to show what's possible. Hopefully, with feedback, we can find a nice equilibrium.

### Some notes of this implementation
1. this only logs PhysiCell output, not BioFVM (namespace concerns, particularly at startup)
2. create a standard way to write a PhysiCell warning/error (`log_warning` and `log_error` functions)
3. an optional variable to `log_output` allows the message to also be printed to `std::cout` so it is easy for devs to have both!
4. `log_warning` and `log_error` default to printing to console
5. I don't know how to easily (read: in one line) convert some of the syntax for `log_output(std::string message)` function signature. Specifically, something like `std::cout << a_string_var << " and " << a_char_var << std::endl;` so I converted these to `log_file << a_string_var << " and " << a_char_var << std::endl;` which works but creates inconsistencies
6. Test it on the template project or else **update your Makefile**

### DIFF
--- a/core/PhysiCell_cell.cpp
+++ b/core/PhysiCell_cell.cpp
@@ -1609,7 +1609,7 @@ void prebuild_cell_definition_index_maps( void )
 		int ID = node.attribute( "ID" ).as_int();  
 		std::string type_name = node.attribute( "name" ).value();   
 
-		std::cout << "Pre-processing type " << ID << " named " << type_name << std::endl; 
+		log_output( "Pre-processing type " + std::to_string(ID) + " named " + type_name );
 		
 //		cell_definitions_by_name[ type_name ] = pCD; 
 //		cell_definitions_by_type[ pCD->type ] = pCD; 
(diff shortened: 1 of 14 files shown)


rheiland:
Thanks for starting this! I spent a few hours playing with all this and want to offer some feedback. First, I very much agree that we need a way to remind/warn users about their model's use/non-use of rules. Recently, I spent ~30 mins scratching my head over why my continued attempts to introduce new rules into a model didn't seem to produce the expected results. Turns out, I had the "enable" checkbox unchecked (on the Rules tab of the Studio). I had been experimenting with using/not using rules during my sims and simply lost track.

First, we should just note that one can perform "simple logging" by redirecting terminal output to a file, e.g., `project > my_sim.log`, at least on a *nix shell. But I don't know if this works on Windows various command line shells.

After spending some time playing around with your approach and then my modified approach (discussed below), I think the output from `display_simulation_status`, i.e., the 
```
current simulated time: 60 min (max: 7200 min)
total agents: 5
interval wall time: 0 days, 0 hours, 0 minutes, and 1.01192 seconds 
total wall time: 0 days, 0 hours, 0 minutes, and 1.01193 seconds 
```
should also include a reminder/warning if rules are *not* being used, e.g.:
```
*** NOTE: rulesets are not being used
```
And I think I'd prefer not including this line if rules are being used. (I also confess ignorance about having multiple `<ruleset ... ` inside the `<rulesets>` block, but each `<ruleset>` (singular) has its own boolean `enabled` attribute, so that makes this reminder/warning more complicated.)

I like the idea of providing (optional) log files. In fact, I convinced myself that it would be good to allow for multiple log files: signals-behaviors, rules, model summary, and simulation output at every `next_full_save_time`. I created a `PhysiCell_logs.xml` file that would be in the root dir (not /config) which simply contained flags for each:
```
<PhysiCell_logs>
    <log_signals_behaviors enable="false">signals_behaviors.log</log_signals_behaviors>
    <log_rules enable="false">rules.log</log_rules>
    <log_model_summary enable="false">model.log</log_model_summary>
    <log_time_steps enable="false">time_steps.log</log_time_steps>
</PhysiCell_logs>
```
One reason for this is that I typically *like* to see the `display_simulation_status` output when I'm running a sim - it reminds me of how often results are being generated.

On the topic of the `log_output` signature, I think the problem goes away if we define a single `std::ostream` to represent *either* `std::cout` or the log file's `std::ofstream`. It's possible - I tested it for a single log file.

Let's not forget other workflows and how they might affect log files, e.g., parameter explorations and/or replicates (same params but different PRNG seeds). We'd probably want to be able to specify different names for the log files, rather than overwriting the same files. Or maybe just sending all output to `/dev/null` (or equivalent).

I appreciate your respect for modularity and the creation of the new `PhysiCell_log.{h,cpp}` files, but I wonder if we should consider simply including this functionality in existing files, e.g., `PhysiCell_utilities` or other, just to avoid the hassle of updating all sample projects' Makefiles. I'm indifferent.

Finally, maybe one important test of logging is to force a segfault during a sim. What is saved in the log(s)? When does flushing happen? Is it the same on all OSes?

drbergman:
if we ever pick this back up, we should include an `<options>` element to control where output goes/verbosity. for now, just toggling between current functionality and that proposed here.

PR #264, opened 2024-07-14
handle duplicate density and parameter

elmbeech:
pull request for the staled pull request #250 and #232

this pull request keeps the changes to the existing code bases as small as possible.
default settings were chosen so that the current models not will break through these changes.
each case was manually tested, the code works as expected. 

add_density and add_parameter functions:
+ Microenvironment::**add_density**: throw error, if a density with the same name already exists.
+ Parameters::**add_parameter** :  throw error, if a parameter with the same name and type already exists.

update_density and update_parameter functions:
+ Microenvironment::**update_density** : new function which does not throw an error but updates the values, if a density with the same name already exists.
+ Parameters::**update_parameter** :  new function which does not throw an error but updates the values, if a parameter with the same name and type already exists.

function calls arguments:
+ **load_PhysiCell_config_file**: additional bool update_variables argument with default set to false.
+ **setup_microenvironment_from_XML** (called from load_PhysiCell_config_file):  additional bool update_density argument with default set to false. 
+ User_Parameters::**read_from_pugixml** (called from load_PhysiCell_config_file): additional bool update_parameter argument with default set to false.

### DIFF
--- a/BioFVM/BioFVM_microenvironment.cpp
+++ b/BioFVM/BioFVM_microenvironment.cpp
@@ -474,8 +474,15 @@ void Microenvironment::resize_densities( int new_size )
 void Microenvironment::add_density( void )
 {
 	// fix in PhysiCell preview November 2017 
-	// default_microenvironment_options.use_oxygen_as_first_field = false; 
-	
+	// default_microenvironment_options.use_oxygen_as_first_field = false;
+
+	// check if density exist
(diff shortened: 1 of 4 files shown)


elmbeech:
superseded by pull request https://github.com/MathCancer/PhysiCell/pull/298.
closed.

PR #266, opened 2024-07-14
add reset function for mesh.mat file generation. 

elmbeech:
+ move the **static bool BioFVM_substrates_initialized_in_dom** variable out of the BioFVM::add_BioFVM_substrates_to_open_xml_pugi function. 
+ add a **BioFVM::reset_BioFVM_substrates_initialized_in_dom** function to be able to reset the BioFVM_substrates_initialized_in_dom variable.


MathCancer:
beautiful. we agree!

Issue #269, opened 2024-07-14
custom cell division; cell lineage 


rheiland:
Rather than submit a PR for this, as it's probably premature, I wanted to share/discuss a branch (https://github.com/rheiland/PhysiCell/tree/custom-division)  that implements:
* a new custom function, `cell_division_function`, that gets invoked when a cell divides
* cell attributes `generation` and `parentID` for cell lineage tracking

To test, rather than add another directory in `sample_projects` (which requires editing all of those Makefiles), I opted to add a `sample_division_lineage` to `user_projects`. Yes, I know this sort of violates its original purpose, but thought we could discuss this practice (with the `sample_` prefix). 

So, to test:
* clone/download the branch
```
make -j2
make load PROJ=sample_division_lineage
make
```
Some results from the Studio are shown. Note that the model defines 2 cell types: `default` and `ctype1`. At the beginning, only 2 `default` cells are created (in cells.csv). When the custom `cell_division_function` for this model is invoked (from `divide()`), it does a coin flip to decide if a daughter cell should differentiate. The 2 cell types have different phenotypes - the `ctype1` chemotaxes to oxygen; `default` does not.

The following image shows results of being able to do asymmetric division, using the new custom function (rf. custom.cpp). However, note that it requires providing the custom function in `create_cell_types` (in custom.cpp); rf. https://github.com/MathCancer/PhysiCell/issues/265 . 
<img width="356" alt="Screenshot 2024-07-14 at 6 18 13 PM" src="https://github.com/user-attachments/assets/e98405ca-ef0a-4356-b395-8927afb17557">

And these screenshots show the outcome of displaying the new cells' `generation` and `parentID` scalars.
<img width="887" alt="Screenshot 2024-07-14 at 5 55 11 PM" src="https://github.com/user-attachments/assets/08995322-64f4-482a-b575-8b348e0c50bc">

drbergman:
This looks amazing! Still plenty to be done to make standard asym div functions, etc, but this will hopefully get us over the barrier to start really working on this. One comment:
1. I think we may want `generation` to increment for *both* daughter cells on division. This number, as I understand it, tracks how many cell divisions from the initial ancestor of this cell until this one. This will not always be recoverable from counting `parentID`'s since some of those may be removed from the simulation before a save event.

rheiland:
[...]

PR #273, opened 2024-08-15
Custom division

rheiland:
Provide functionality for custom cell division (a new `functions.cell_division_function`), plus lineage tracking via `generation` and `parentID` attributes for agents.


modules/PhysiCell_MultiCellDS.cpp
```diff
@@ -890,6 +890,29 @@ void add_PhysiCell_cells_to_open_xml_pugi_v2( pugi::xml_document& xml_dom, std::
 		data_start_indices.push_back( index ); 
 		cell_data_size += size; 
 		index += size; 
+
+        // generation
+		name = "generation";
```
drbergman, line 895: when merging with dev-paul, will need (or maybe just want?) to update these with Paul's condensed version of this code there

drbergman (commented):
Really excited for this! There's also some changes on where the PhysiCell_constants are defined. Is that related? Or solving some other issue?

rheiland:
> Really excited for this! There's also some changes on where the PhysiCell_constants are defined. Is that related? Or solving some other issue?

This is one reason I've been having mild panic attacks lately :-) over the upcoming merge party. I think I made this branch from my branch which dealt with https://github.com/MathCancer/PhysiCell/pull/248 (I had reasons for doing so at the time). So this touches on merging PRs in a certain order, I think. Or, I could simply update the branch for custom cell division and use the original "constants" code.

rheiland:
Earlier comments about this PR are in the related Issue https://github.com/MathCancer/PhysiCell/issues/269

rheiland:
Closing. The actual custom function for cell division was merged with #285. Perhaps later we can refer to this 273 for an improved cell lineage tracking PR.

PR #275, opened 2024-08-16
make output folder if it does not exist

drbergman:


drbergman:
cleaned up by and superseded by #309

PR #279, opened 2024-08-18
comment out unused calculation

rheiland:
In `add_potentials`, the `RN` calculation is no longer used.


MathCancer:
fixed elsewhere, thank you!

PR #280, opened 2024-08-19
Rebased development-paul + Randy's PR

vincent-noel:
So I just did 

```
git checkout -b development-paul mathcancer/development-paul
git rebase mathcancer/development
```

and it went smoothly. Sometimes we have conflicts to fix, but not in this case. 

and then
```
git cherry-pick c629ed5
```
to add Randy's fix. 

Interestingly, some tests are failing :D I turned off the option to cancel all tests if one fails, so we can look exactly at which tests are failing, and why. Most likely it's not important, but we need to check. I have to run an errand, I'll look at it in a few hours.


vincent-noel:
Now it should be ready !

- I fixed the interaction sample model (replaced state.damage with phenotype.cell_integrity.damage).
- I introduced some changes to the script to run the tests, in order to reduce stochasticity (deterministic cell cycle & death phases, deterministic migration, no dynamics spring attachments). 
- I removed some models (heterogeneity, interactions,  rules and cancer_invasion) which despite the stochasticity reduction, still have enough stochasticity to fail the tests when we add PRNG draws.

drbergman (changes requested):
I see some files that maybe should not have been changed? For example, there's a new out.gif that we probably don't want to commit. Also, there's a rules.csv file renamed to blah.txt in one of the PhysiBoSS samples. In total,
* PAUL_STUFF.zip
* sample_projects_intracellular/boolean/cancer_invasion/config/rules.csv -> config/test_dir/blah.txt
  * we'll want to be careful populating the config folder especially now that we have the recursive cp
* dictionaries.txt
* out.gif

core/PhysiCell_standard_models.cpp
```diff
@@ -1231,13 +1259,49 @@ void standard_cell_cell_interactions( Cell* pCell, Phenotype& phenotype, double
 			double attack_ij = phenotype.cell_interactions.attack_rate(type_name); 
 			double immunogenicity_ji = pTarget->phenotype.cell_interactions.immunogenicity(pCell->type_name); 
 
+			// probability of STARTING an attack 
 			probability = attack_ij * immunogenicity_ji * dt; 
-			
-			if( UniformRandom() < probability && attacked == false ) 
+			if( attacked == false && pCell->phenotype.cell_interactions.pAttackTarget == NULL )
```
drbergman, line 1264: @MathCancer I think we may want to initialize `attacked` above as 
```
bool attacked = pCell->phenotype.cell_interactions.pAttackTarget != NULL
```
Otherwise, here we are allowing a cell that is continuing to attack from a previously initiated attack to start attacking a new target. that's confusing so here's what could happen:

1. At $t=t_1$, T cell $T$ initiates an attack against cancer cell $C_1$. During $t=t_1$ update, $T$ cannot start a separate attack against any other cell
2. At time $t=t_2>t_1$, $T$ is still attacking $C_1$. It chooses to initiate an attack against a different cancer cell, $C_2$ so that now it is attacking two cells.
I think some bad things could happen, e.g., $T$ is now targeting $C_2$ but still spring-attached to $C_1$.

MathCancer:
I can do cleanup as we get close tothe actual release. :-)

MathCancer:
> I see some files that maybe should not have been changed? For example, there's a new out.gif that we probably don't want to commit. Also, there's a rules.csv file renamed to blah.txt in one of the PhysiBoSS samples. In total,
> 
> * PAUL_STUFF.zip
> * sample_projects_intracellular/boolean/cancer_invasion/config/rules.csv -> config/test_dir/blah.txt
>   
>   * we'll want to be careful populating the config folder especially now that we have the recursive cp
> * dictionaries.txt
> * out.gif
> * todo.txt
> 
> fin

I might need those notes until we get it all done. PAUL_STUFF I believe has a sample project/ test taht I''ll wnat ot adapt to show the newer T cell and phagocytosis routines.

PR #281, opened 2024-08-20
update samples to not need random_seed in userpars

drbergman:
Update cancer_immune_3D, ecoli_acetic_switch, and substrate_internalization (in unit tests)

Not sure how I missed these in my first sweep through the custom files...


MathCancer:
As described, this is well-thought-out. Let's accept.

PR #282, opened 2024-08-26
Fix random seed warning message

drbergman:
I want to still default to using the system clock, which means doing that even if it's not set in the options. This will give a more actionable error than if a user removes BOTH the options element and the user parameter.


MathCancer:
Makes sense to accept alongside #281

Issue #283, opened 2024-08-26
volume NOT conserved on cell transformation; also duplicate Volume class


drbergman:
When a cell transforms to another cell type, the code (v1.13.1) sets the volume of the cell to the volume of the new cell definition. Here are two snapshots demonstrating this in which a `small` cell type (grey) transforms to a `big` cell type (red) and immediately changes volume. **Note:** the two snapshots are 6 min apart.
<img width="300" alt="image" src="https://github.com/user-attachments/assets/a23f0988-e01c-484d-b76b-48461ac0edbd"><img width="300" alt="image" src="https://github.com/user-attachments/assets/a4012570-449a-42ef-a2bb-2297ed7f8b5c">

### Proposed solution
Within the function `Cell::convert_to_cell_definition` called by `standard_cell_transformations`, the cell has its phenotype overwritten by `phenotype = cd.phenotype;`. As `Volume` is a member of `Phenotype`, this is where the volume is changed. **I propose we change only the volume targets here, not the actual volume values** (`total`, `solid`, `fluid`, etc). Here's how that would look:
```
void Cell::convert_to_cell_definition( Cell_Definition& cd )
{	
	Volume cell_volume = phenotype.volume;
	...
	phenotype = cd.phenotype; 
	phenotype.volume = cell_volume; // leave the cell volume alone..
	// ..except for the following target values
	// phenotype.volume.target_solid_cytoplasmic = cd.phenotype.volume.target_solid_cytoplasmic; // gets set by standard_volume_update_function anyways
	phenotype.volume.target_solid_nuclear = cd.phenotype.volume.target_solid_nuclear;
	phenotype.volume.target_fluid_fraction = cd.phenotype.volume.target_fluid_fraction;
	phenotype.volume.target_cytoplasmic_to_nuclear_ratio = cd.phenotype.volume.target_cytoplasmic_to_nuclear_ratio;
	phenotype.volume.relative_rupture_volume = cd.phenotype.volume.relative_rupture_volume;
	...
```
After these lines, `set_total_volume( phenotype.volume.total )` is called. That would be unnecessary now since the state values for volume would not be changing. The part that would need to be called is the line that updates the `max_cell_interactive_distance_in_voxel` as the new cell def might have a bigger interaction distance.

### Also there's a duplicate `Volume` class?
[...]

PR #284, opened 2024-08-26
Preserve volume when converting to cell definition

drbergman:
Only change target volumes when convert_to_cell_definition is called. This will ensure that cell transformations do not cause instantaneous changes in cell volume.

This is responsive to #283 (except the duplicate `Volume` class bit)


MathCancer:
Let's do it!

MathCancer:
this all looks great!@

PR #286, opened 2024-08-28
Upgraded gcc to v13 in macos binaries builds

vincent-noel:
Quick fix upgrading the version of gcc used to build the mac binaries (mostly needed for M1s, but also upgraded on x86 macs since it will probably help with compatibility).

### DIFF
--- a/.github/workflows/build_binaries.yml
+++ b/.github/workflows/build_binaries.yml
@@ -130,7 +130,7 @@ jobs:
     - uses: actions/checkout@v4
 
     - name: Install dependencies
-      run : brew install gcc@11
+      run : brew install gcc@13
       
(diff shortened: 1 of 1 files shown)


drbergman:
This might be useful to always get the latest gcc compiler: https://github.com/marketplace/actions/install-gcc

drbergman:
There's an open issue on supporting gcc13. sounds like it works now. someone suggested a different package manager: https://github.com/egor-tensin/setup-gcc/issues/12#issuecomment-1758706315

vincent-noel:
Yeah, better do it manually, it's mainly for macos, so it's just a brew instal gcc@13.

drbergman (approved):

PR #287, opened 2024-08-28
remove unwanted files; fix cancer_biorobots bug; etc.

rheiland:
Removed PAUL_STUFF.zip, out.gif, .travis.yml, and todo.txt

PR #290, opened 2024-09-03
Fix bug in matlab/read_MultiCellDS_xml.m

drbergman:
Responsive to issue raised in slack re bugs in using the `matlab/read_MultiCellDS_xml.m` script. Perhaps arose due to changes in multicellds output without synching the matlab implementation. see below for details

Bugs handled:
* only one `simplified_data` tag. don't look for another
* `MAT.basic_agents` --> `MAT.cells`

Other changes:
* refactored `matlab/read_MultiCellDS_xml.m` to use local functions
* some renaming of nodes to make more readable
* use `str2double` for efficiency (but really to make matlab not cry)
* eliminate unused variables for efficiency (but also for the not crying thing)
* `== false` --> `~`
* more efficient `setfield` implemented (no matlab, no cry)
* includes `MCDS_constants` in `MCDS.constants`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+*.asv
 ._.DS_Store
 .DS_Store
 *.exe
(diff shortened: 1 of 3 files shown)


drbergman:
New other change:
* add `custom-division-sample` to Makefile

rheiland:
> New other change:
> 
> * add `custom-division-sample` to Makefile

Duplicate of this I think, but never hurts :)   https://github.com/MathCancer/PhysiCell/pull/287/commits/1f4da4ca699219fd8611ab577768b064dac22e96

PR #292, opened 2024-09-07
apply rules BEFORE checking for transformation

drbergman:
Everything else (in phenotype updates) is after applying the rules. with it ordered previously (transform then apply rules), it is possible for cells at the beginning of a sim to transform before the rule inhibiting it is applied.

### DIFF
--- a/core/PhysiCell_cell.cpp
+++ b/core/PhysiCell_cell.cpp
@@ -1170,6 +1170,9 @@ void Cell::convert_to_cell_definition( Cell_Definition& cd )
                 * phenotype.mechanics.relative_maximum_adhesion_distance;
         }
 	}
+
+	if( PhysiCell_settings.rules_enabled )
+	{ apply_ruleset( this ); } // don't wait for the start of the next phenotype update to apply the rules to this cell
 	return; 
 }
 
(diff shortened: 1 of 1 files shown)


drbergman:
can wait until after v1.14.0

drbergman:
Example demonstrating need:

`content` cells begin in a well-oxygenated environment. A rule says that low oxygen increases their transformation rate to `migratory`, a migratory phenotype, defined by a different cell type to reflect that this change is not transient. The rule is therefore:
```
content,oxygen,decreases,transform to migratory,0,5,4,0
```
and acts on a baseline transformation rate of 0.1 `content`->`migratory`.

### current version result
The current version of PhysiCell does the 0.1 transformation rate **before** applying the rules, including in the first iteration. Thus, some of these cells (about 0.1 * 6 = 60%) transform in this first time step as seen here:
imageimageimage

### post-merge result
Now, we will first apply the rulesets within the phenotype update *before* transformations. Any call to `convert_to_cell_definition` will apply the rules to the cell after it has assumed its new type. Now we see the expected result that all cells remain `content` at the start of the simulation:
imageimageimage

drbergman:
a similar argument applies to when we compute the gradient vectors. those are calculated first only after the first application of the rules. In this case, the gradient calculation should happen before the rules are applied to avoid this kind of behavior.

Note: this PR does not address this gradient-based issue

PR #298, opened 2024-09-21
add update_density  and update_parameter function

elmbeech:
pull request for the staled pull request https://github.com/MathCancer/PhysiCell/pull/264, https://github.com/MathCancer/PhysiCell/pull/250 and https://github.com/MathCancer/PhysiCell/pull/232.

this pull request is compatible with physicell version 1.14.0 and keeps the changes to the existing code bases as small as possible.
default settings were chosen so that the current models not will break through these changes.
each case was manually tested, the code works as expected.

add_density and add_parameter functions:

    Microenvironment::add_density: throw error, if a density with the same name already exists.
    Parameters::add_parameter : throw error, if a parameter with the same name and type already exists.

update_density and update_parameter functions:

    Microenvironment::update_density : new function which does not throw an error but updates the values, if a density with the same name already exists.
    Parameters::update_parameter : new function which does not throw an error but updates the values, if a parameter with the same name and type already exists.

function calls arguments:

    load_PhysiCell_config_file: additional bool update_variables argument with default set to false.
    setup_microenvironment_from_XML (called from load_PhysiCell_config_file): additional bool update_density argument with default set to false.
    User_Parameters::read_from_pugixml (called from load_PhysiCell_config_file): additional bool update_parameter argument with default set to false.


elmbeech:
as discussed, update_density and update_parameter through an error, if one tries to update a density or parameter that not already exists. the rest is the same.

modules/PhysiCell_settings.cpp
```diff
@@ -523,6 +523,65 @@ void Parameters<T>::assert_not_exists( std::string search_name )
...
+	}
+
+	// change value
+	parameter_index = it->second;
+	parameters[parameter_index].value = my_value;
+
+	return;
+}
+
+template <class T>
```
drbergman, line 545: Can you call the above implementation of `update_parameter` to simplify this function and the codebase?

BioFVM/BioFVM_microenvironment.cpp
```diff
@@ -487,76 +487,112 @@ void Microenvironment::add_density( std::string name , std::string units )
...
+		std::cout << "Error: density named " << name << " does not exist. Cannot update density!" << std::endl;
+		exit(-1);
+	}
+
+	// update units
+	density_units[density_index] = units;
+	return;
+}
+
+void Microenvironment::update_density( std::string name , std::string units, double diffusion_constant, double decay_rate )
```
drbergman, line 580: Can you use the above implementation of `update_density` to simplify this function and the codebase?

drbergman (changes requested):

modules/PhysiCell_settings.cpp
elmbeech: @drbergman could you explain to me how this is done?
Maybe because I am not too familiar with templates.
Could you please give me an example of what you wish for?
Then I will make the changes accordingly.

elmbeech: @drbergman  im i not sure if i really can do what you ask for, as  i try to update parameters.
i don't have "default values",  as in the add case.
this is as concise as i can do it.
please write here how i really should do it so that it is acceptable for you. thank you.

image

drbergman: My intention is something super simple and it seems you took a big step towards it having all 3 of these call `assert_exists`. But something like this could make it even tighter:

```
void Parameters<T>::update_parameter( std::string my_name , T my_value , std::string my_units )
{
	update_parameters( my_name, my_value); // will check my_name exists and then set the value to my_value
	parameters[find_index(my_name)].units = my_units;
	return;
}
```

drbergman: or I think you could even use `parameters[my_name]` instead of `parameters[find_index(my_name)]`

drbergman: also, looking at the `[]` operator on strings, it already checks for existence.

elmbeech: i can do that, although i think this will make the code slower.

drbergman: oh yeah? I mean, we could throw in some `const` and pass by reference to reduce any overhead costs. I imagine these functions won't be called nearly enough though to make a noticeable impact on performance. Or will they?

Finally, I personally find this kind of coding to be so much easier to maintain that it easily makes it worth the computational tradeoff (at least as I imagine the magnitude of the cost)

elmbeech: is this ok?
image

drbergman: hmmm...I don't think I'm able to communicate this...your original is the best of these, in my opinion

elmbeech: oh yes, it will get called often. I my case every iteration.
but i know, the PhysiCell core code is not written for only one project. so this argument has no weight.

BioFVM/BioFVM_microenvironment.cpp
elmbeech: is this ok?

image

drbergman: yup

elmbeech: i still think it is odd to have to implement a slower version because of code maintenance. 
before we had to call the find_index function 1 time, now 3 or 4 times. 
just to avoid storing the value in a variable?

elmbeech: this is the compromise i suggest between maintenance (as closed as it get to the existing add_parameter function) and speed. i hope this is acceptable.

image

elmbeech: and the same for the densities:

image

elmbeech: i pushed these changes and hope they are ok and get merged.

elmbeech: @drbergman could you please confirm that the conflict / conversation has resolved (as i think so ).
or let me know if otherwise?

thank you, Elmar

modules/PhysiCell_settings.cpp
elmbeech: @drbergman could you please confirm that the conflict / conversation / change you requested has resolved (as i think so ).
or let me know if otherwise?

i don't think it is appropriate when i click the resolved button.

thank you, Elmar

elmbeech:
@drbergman can you please give your ok to this change?
I really would like to get this pull request in now!
thank you, Elmar

elmbeech:
Hi Paul! 
I attached a zip folder which contains `physicell 1.14.1` inclusive the changes from `this pull request`.
The `main.cpp` and the `custom.cpp` were adjusted to run consecutive episodes.
+ the `main.cpp` contains an extra `episode loop`.
+ the `custom.cpp` has `generate_cell_types` and a `reset_cell_types` function, which are based on the original `create_cell_types function`. 
I tried to use the physicell high-level functions as much as possible.
All works well, if episodes run for e.g. 1440 min, but if episodes are e.g. set to 7200 min, the in the following episode I run into a core dump, when I try to delete all the cells. 
This is where I am at this moment stuck with this pull request.
Elmar
[Dec_26_2024_0443.zip](https://github.com/user-attachments/files/18252042/Dec_26_2024_0443.zip)

rheiland:
I'd suggest making any new arguments to functions as optional, if possible. E.g., `void initialize_microenvironment( bool update_microenvironment = true)`. This would avoid needing to edit all the sample projects.

elmbeech:
Thank you, @rheiland! 
This is exactly the way I implemented it.
BioFVM/BioFVM_microenvironment.cpp line 1178 (1178 -1230) and BioFVM/BioFVM_microenvironment.h line 370.
However, I set the default form bool update_microenvironment = false because update is only needed if you run another episode. 
The code is passing all the old unit tests code.

elmbeech:
superseded by pull request: https://github.com/MathCancer/PhysiCell/pull/344

modules/PhysiCell_settings.cpp
```diff
@@ -658,6 +711,8 @@ bool setup_microenvironment_from_XML( pugi::xml_node root_node )
 		// add the substrate 
 		if( i == 0 )
 		{ microenvironment.set_density( 0, name, units ); }
+		else if( update_density )
+		{ microenvironment.update_density( name, units ); }
```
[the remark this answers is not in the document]
elmbeech: this is correct.
i went with a scaled down version in a follow up pr and closed this pr.

PR #308, opened 2024-10-14
Conserved quantities are transferred by default during phagocytosis

MathCancer:
In response to Issue https://github.com/MathCancer/PhysiCell/pull/307, cells by default ingest 100% of internalized substrates and conserved variables when eating another cell.


core/PhysiCell_cell.cpp
```diff
@@ -1399,12 +1399,33 @@ void Cell::ingest_cell( Cell* pCell_to_eat )
...
+
 		*internalized_substrates += *(pCell_to_eat->internalized_substrates); 
 		static int n_substrates = internalized_substrates->size(); 
 		pCell_to_eat->internalized_substrates->assign( n_substrates , 0.0 ); 	
+
+		// conserved quantitites in custom data aer divided in half
+		// so that each daughter cell gets half of the original ;
+		for( int nn = 0 ; nn < custom_data.variables.size() ; nn++ )
+		{
+			custom_data.variables[nn].conserved_quantity = true;
```
drbergman, line 1414: I think we want to delete this line.

drbergman (changes requested):

core/PhysiCell_cell.cpp
MathCancer: oof, right.  I think I was forcing it during testing since it wasn't being parsed.

MathCancer: did the change, thank you!

heberlr:
I changed the comment about the conservation for phagocytosis. Sorry about that second commit because I accidentally changed two lines in the first one.

drbergman (approved):

PR #314, opened 2024-10-23
handle abspath for output folder creation

drbergman:
any path starting with the path separator (indicating an absolute path) will cause an error with the current implementation _even if the path already exists_. This is because it attempts to create a directory with the empty string. This handles checking for that case. It splits on OS to check the path separator. For windows, it checks for the abspath starting with the drive. If we can upgrade to c++17, we could take advantage of `std::filesystem` to greatly simplify this. something like 

```
// Function that we could use if we used c++17
// THIS IS NOT WHAT IS IMPLEMENTED IN THIS PR
bool create_directories(const std::string &path)
{
    std::error_code ec;
    bool success = std::filesystem::create_directories(path, ec);
    if (!success && ec) {
        // Handle error or log it, if necessary
        return false;
    }
    return true;
}
```

### DIFF
--- a/modules/PhysiCell_settings.cpp
+++ b/modules/PhysiCell_settings.cpp
@@ -329,15 +329,14 @@ bool create_directories(const std::string &path)
     std::string currentPath;
 
 	// Check for Unix-like absolute path or Windows absolute path with drive letter
-#if defined(__MINGW32__) || defined(__MINGW64__)
-	if (path[0] == '\\')
-	{ pos = 1; } // Windows absolute path starting with backslash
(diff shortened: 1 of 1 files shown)


rheiland:
On my Windows machine, using:
```
PS C:\Users\randy\dev\PhysiCell-fix-create-output-from-abspath\PhysiCell-fix-create-output-from-abspath> g++ --version
g++.exe (Rev6, Built by MSYS2 project) 13.1.0
```
I got an error if my absolute path started with a "/". However, if I prefixed it with the drive: "c:/" then it worked, e.g.:
```
<folder>c:/Users/randy/dev/PhysiCell-fix-create-output-from-abspath/PhysiCell-fix-create-output-from-abspath/out2</folder>
```

PR #315, opened 2024-10-24
initialize total_attack_time to 0.0

drbergman:
Previously, it was unset and so it seems one cell (the first placed?) would have an arbitrary value.
imageimage


drbergman:
Or maybe we wanted to move this to `Cell_Integrity` along with `damage`?

PR #317, opened 2024-11-06
fix buffer bug in writing coords in xml

drbergman:
sidestep issue of overflowing buffer in writing the coordinates to the xmls by using `std::ostringstream`. Note: PhysiBoSS does use this already in `BooleanNetwork.h`

### DIFF
--- a/BioFVM/BioFVM_MultiCellDS.cpp
+++ b/BioFVM/BioFVM_MultiCellDS.cpp
@@ -570,36 +570,9 @@ void add_BioFVM_substrates_to_open_xml_pugi( pugi::xml_document& xml_dom , std::
 		// if Cartesian, add the x, y, and z coordinates 
 		if( M.mesh.Cartesian_mesh == true )
 		{
-			char temp [10240];
-			int position = 0; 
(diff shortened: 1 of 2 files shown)


rheiland:
I've reviewed and approve this PR. This is an important fix for handling large simulation domains (in fact, one user already reported the bug). It avoids the hard-coded array for writing out the domain coords into the XML.
* using ostringstream only requires C++-11
* any downsides to using ostringstream don't seem too worrisome (perhaps slower for VERY large strings which should not apply here)
* tested on Mac and Windows - compiles and runs without a problem; visualized results  in the Studio
* if we wanted to take it even further, we could include exception handling (try-catch in `write_coordinates_node`), but I'm OK as it is. I even tested using exception handling, but it didn't help gracefully catch memory corruption (just gave a segfault).
* note that there's another hard-coded array in that section of code, but it seems safer at least.

vincent-noel (approved):
Looks good to me too !
I was curious at first about the exact format used for doubles here: we specify %f with the sprintf, but nothing with the stream. But as far as I understood, this is indeed the default with the stream, and both have 6 digits precision by default, so it should be exactly the same. 
 Thanks Daniel!

PR #318, opened 2024-11-06
fix bug in export rules

drbergman:
- record the value of applies_to_dead
- ~~use v3 for export (to make sharing easier)~~ do not export the rules (removed in commit [9820c77](https://github.com/MathCancer/PhysiCell/pull/318/commits/9820c77980ce8f69636d68a44d4b0bb5c68ff4d2))
- still create the .txt and .html rules files
- copy the rules file(s) (the CSVs actually read in) to the output directory

### DIFF
--- a/core/PhysiCell_rules.cpp
+++ b/core/PhysiCell_rules.cpp
@@ -2193,7 +2193,7 @@ void export_rules_csv_v1( std::string filename )
 				{ response = "decreases"; max_response = min_value; }
 				double half_max = pHRS->rules[k]->half_maxes[i];
 				double hill_power = pHRS->rules[k]->hill_powers[i];
-				bool use_for_dead = false; 
+				bool use_for_dead = pHRS->rules[k]->applies_to_dead_cells[i];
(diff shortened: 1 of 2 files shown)


drbergman:
we could add on to this PR the copying of the input rules file into the output folder. it may conflict with the file we currently write here, so we'd have to decide whether to keep this file or how we will name these two files.

MathCancer:
Hello!

I disagree with "copying rules instead of writing new one". Suppose that PhysiCell fails to parse a rule but continues to run. If the rules file is simply copied, then the user thinks that the model that actually ran matches the input file. 

This is why, in part, I have PhysiCell write rules in the output folder, based upon the rules that successfully were parsed and are being executed. Then the rules file in output matches the actual rules of the simulation.

I would propose that we update the export to match current rules format, rather than switch to a raw copy. (Or do both.) 

I do agree that "applies to dead" needs to be annotated.

MathCancer (approved):
looks terrific!

Issue #321, opened 2024-11-10
save random seed from <random_seed>system_clock</random_seed>


rheiland:
When a model specifies `system_clock` to provide the RNG seed, it would be nice to have the seed value saved for later attempts at reproducibility (primarily if a single thread is used).

drbergman:
100% agreed.

drbergman:
Are we ok just writing a new file `output/random_seed.txt` that holds it? Otherwise, this will require a bit more restructuring to achieve.

PR #322, opened 2024-11-11
write random_seed to output/random_seed.txt

drbergman:
Responsive to #321


rheiland (approved):
Tested on Windows and works. So now we have the random seed written to stdout along with other sim info, plus saved in a .txt file in the output dir. Thanks!

PR #330, opened 2024-11-15
Removed call to SeedRandom() in main of physiboss cell lines project

vincent-noel:
@rheiland caught this today, this is the fix !


drbergman (commented):
Makes sense. The `create_cell_types();` call on L129 of `main.cpp` also (potentially) sets the random seed. It depends on if `random_seed` is a user parameter. If you want to keep that option open for your users, then leave as is. Otherwise, for this project...
1. remove L78-83 in the `custom.cpp`
2. remove `random_seed` from `user_parameters` in the config
3. add the following inside the `<options>` element of the config (starts at L44)
```
<random_seed>0</random_seed>
```

PR #332, opened 2024-11-15
remove macos-12 from ci

drbergman:
gh actions deprecation [upcoming](https://github.com/actions/runner-images/issues/10721)
still have macos-14 ci

### DIFF
--- a/.github/workflows/build_binaries.yml
+++ b/.github/workflows/build_binaries.yml
@@ -124,29 +124,34 @@ jobs:
           {short_name: "interaction", project: "interaction-sample", name: "PhysiCell Interaction", binary: "interaction_demo", extra_run: ""},
        ]
     
-    runs-on: macos-12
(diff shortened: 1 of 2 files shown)


vincent-noel (changes requested):
Ah yes, I forgot about this. We are getting close. Thanks for the reminder :)

We shouldn't just remove it, because 12 and 13 are still x86_64 (non arm64 or M1) chips, and we should test on these as long as we can. So best would be to just change every macos-12 to macos-13, and pray that it works. 

Same thing for the build_binaries script, which is building the universal (x86 and arm86) binaries for mac, for this we still need a x86_64 environment. 

If you have time to do this, also for the build_binaries.yml, that's awesome, otherwise I can do it hopefully next week !

drbergman:
done! now we just hold our breath for the tests...

PR #336, opened 2024-11-21
copying rules file(s) instead of writing new one

drbergman:


drbergman:
we should just use #318

Issue #320, opened 2024-11-10
rare nullptr in get_single_signal


rheiland:
In a particular model which @drbergman and @heberlr have been running replicates on, there seem to be occasional runtime errors. In the current development branch, at this line:
https://github.com/MathCancer/PhysiCell/blob/development/core/PhysiCell_signal_behavior.cpp#L1049 
if we insert this code (after `dead_cells++;`), it seems to sometimes detect a nullptr and at least avoid a fatal error:
```
                Phase* myPhase = &(pC->phenotype.cycle.current_phase());
                if ( myPhase == nullptr)
                {
                      std::cout << __FUNCTION__ << "  -------- myPhase = nullptr\n";
		      return -9e9; 
                }
```
Still unsure of the underlying cause, but maybe others have a clue. In my initial tests, it also seems to be more common to happen after the final call to `update_all_cells` in main.cpp, before leaving the `while` loop and saving the `final.*` simulation outputs.

rheiland:
This error seems to be so rare and weird and not reproducible that I'm closing this issue for now. But here's the latest error that I saw while running lldb with a clang-compiled executable.
```current simulated time: 1320 min (max: 7200 min)
total agents: 1388
interval wall time: 0 days, 0 hours, 0 minutes, and 14.5646 seconds 
total wall time: 0 days, 0 hours, 5 minutes, and 28.0556 seconds 

error: project_clang_debug debug map object file "/Users/heiland/dev/PhysiCell-development/PhysiCell_signal_behavior.o" changed (actual: 0x6733df39, debug map: 0x672f7fda) since this executable was linked, debug info will not be loaded
error: project_clang_debug debug map object file "/Users/heiland/dev/PhysiCell-development/PhysiCell_rules.o" changed (actual: 0x6733df38, debug map: 0x672f7fdb) since this executable was linked, debug info will not be loaded
Process 56308 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x4)
    frame #0: 0x00000001000c86c4 project_clang_debug`PhysiCell::get_single_signal(PhysiCell::Cell*, int) + 2396
project_clang_debug`PhysiCell::get_single_signal:
->  0x1000c86c4 <+2396>: ldr    w8, [x8, #0x4]
    0x1000c86c8 <+2400>: adrp   x9, 110
[...]

PR #338, opened 2024-11-24
refactor update_all_cells

drbergman:
This is responsive to issue #320 

- fix bug in neighbor list
    - phagocytosis and fusion could result in removal of a cell
    - but the neighbor list was not updated
- resolve interactions after phenotype updates
    - this allows the update velocity step to update the neighbors list one time
    - does move interactions before several steps:
        - gradient computation
        - evaluate interactions (so fewer interactions to evaluate) 
        - custom rule 
        - update velocity 
        - springs
- ~~do not allow spontaneous spring detachments caused by attack~~ (now a separate pr in #340 )
- refactor how timing is done for simplicity

### DIFF
--- a/core/PhysiCell_cell_container.cpp
+++ b/core/PhysiCell_cell_container.cpp
@@ -124,12 +124,12 @@ void Cell_Container::update_all_cells(double t, double phenotype_dt_ , double me
 {
 	// secretions and uptakes. Syncing with BioFVM is automated. 
 
-	static double time_since_last_cycle = phenotype_dt_;
+	static double time_since_last_phenotype = phenotype_dt_;
 	static double time_since_last_mechanics = mechanics_dt_;
 	static double phenotype_threshold = phenotype_dt_ - 0.5 * diffusion_dt_;
 	static double mechanics_threshold = mechanics_dt_ - 0.5 * diffusion_dt_;
 
-	bool time_for_phenotype = time_since_last_cycle > phenotype_threshold;
+	bool time_for_phenotype = time_since_last_phenotype > phenotype_threshold;
 	bool time_for_mechanics = time_since_last_mechanics > mechanics_threshold;
 
 	#pragma omp parallel for 
@@ -174,7 +174,7 @@ void Cell_Container::update_all_cells(double t, double phenotype_dt_ , double me
 		{
 			if( (*all_cells)[i]->is_out_of_domain == false )
 			{
-				(*all_cells)[i]->advance_bundled_phenotype_functions(time_since_last_cycle);
+				(*all_cells)[i]->advance_bundled_phenotype_functions(time_since_last_phenotype);
 			}
 		}
 		// process divides / removes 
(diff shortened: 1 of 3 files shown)


core/PhysiCell_cell_container.cpp
```diff
@@ -124,6 +124,14 @@ void Cell_Container::update_all_cells(double t, double phenotype_dt_ , double me
 {
 	// secretions and uptakes. Syncing with BioFVM is automated. 
 
+	static double time_since_last_cycle = phenotype_dt_;
```
heberlr, line 127: Awesome! I like this new structure—it's clearer and gets rid of the initialized test. Consider renaming the variable `time_since_last_cycle` to `time_since_last_phenotype` for consistency.

drbergman: Thanks! And done!

core/PhysiCell_cell_container.cpp
```diff
@@ -159,54 +161,58 @@ void Cell_Container::update_all_cells(double t, double phenotype_dt_ , double me
...
+	if( time_for_mechanics )
 	{
-		if(!initialzed)
+		// new March 2022: 
+		// run standard interactions (phagocytosis, attack, fusion) here 
+		#pragma omp parallel for 
+		for( int i=0; i < (*all_cells).size(); i++ )
+		{
+			Cell* pC = (*all_cells)[i]; 
+			standard_cell_cell_interactions(pC,pC->phenotype,time_since_last_mechanics);
```
heberlr, line 209: We need to discuss this change in call order further, as it will impact phagocytosis, attack, and fusion. I noticed it also relates to other PRs (#340 and #341). This change might introduce too many side effects.

drbergman: Yes, very happy to discuss more. By side effects, do you mean unintended behavior and/or errors? Or do you mean changes from how it ran before?

If we find this to be a long discussion, I would prefer to get a quick fix for this bug while we keep working on a longterm fix. Perhaps something like (the highly inefficient) #339?

heberlr: I mean unintended behavior (like in PR #340 ), as the attach and detach dynamics were working fine in the previous release. I feel that by making this call order change, we might introduce these issues and possibly others we haven’t considered yet. Does that make sense?

drbergman: Yeah, that makes sense. To be clear, #340 and #341 are built off the dev branch / previous release (1.14.0), not this fix. Their point of origin is how we implemented changes in 1.14.0. This fix does not change the existence of those behaviors.

MathCancer:
I don't think the correct fix to a bug is to reorder operations to avoid the bug.

core/PhysiCell_cell_container.cpp
MathCancer, line 1: I don't think the correct fix to a bug is to reorder operations to avoid the bug.

MathCancer, line 1: I don't think the correct fix to a bug is to reorder operations to avoid the bug.

drbergman:
Another option that @rheiland worked out was to update the `delete_cell()` function to have it also update neighbors the same way it already handles updating spring attachments

MathCancer:
I do think we need to say: if lingering neighbors after cell deletion is the problem, then the correct fix is to correct those neighbor lists at the time of cell deletion. 

(This is what we did in spring links. We just do the same for neighbor lists now.) 

If A is neighbor to (B,C,D), then B, C and D are also neighbors to A. When A is deleted, remove A from the neighbor lists of B, D and D

drbergman:
Here was where @rheiland documented this fix: https://github.com/MathCancer/PhysiCell/issues/320#issuecomment-2509977089

My main reason for not opting for this solution is that we do not enforce that the `neighbors` is a symmetric relationship. The implementation of `standard_update_cell_velocity` produces a symmetric relationship (by its reliance on `add_potentials`), but we do not impose this restriction on user-defined `update_velocity` functions

drbergman:
another fix is being formulated...

PR #339, opened 2024-11-25
update neighbors to fix phagocytosis bug

### LINKED ISSUE #320: rare nullptr in get_single_signal

drbergman:
This is an alternate way to resolve #320. This is simpler than the way proposed in #338 but I think that one is a better rework.

## Notes
- make sure every cell has an updated neighbors list to start
- this is the simplest change (fewest lines) to fix the bug

## Warning
However, the interactions sample on Ubuntu has a 'std::bad_alloc' what(): std::bad_alloc` error thrown. I cannot reproduce this running on my Mac or a Linux-based HPC. I do not think this fix is triggering that error since none of the other tests are raising that error.

### DIFF
--- a/core/PhysiCell_cell.cpp
+++ b/core/PhysiCell_cell.cpp
@@ -318,6 +318,9 @@ void Cell::advance_bundled_phenotype_functions( double dt_ )
 	// perform transformations 
 	standard_cell_transformations( this,this->phenotype,dt_ ); 
 
+	// make sure neighbors are updated
+	state.neighbors = nearby_cells();
+	
 	// New March 2023 in Version 1.12.0 
 	// call the rules-based code to update the phenotype 
 	if( PhysiCell_settings.rules_enabled )


rheiland:
I did not an error running the `./interaction_demo` on my Ubuntu (Ubuntu 24.04 LTS) with g++ 13.2.0.

rheiland:
When  I run the template `project` in this PR and the one in https://github.com/MathCancer/PhysiCell/pull/338 on the .xml and cell_rules.csv that you shared in slack (with the usual 1 core, fixed seed), I see non-bitwise reproducible results: qualitatively "the same" for snapshot00000001.svg, but not identical. Did you see the same? If so, do you know why?

drbergman:
Making sure we're on the same page: this is a comment about #338 rather than this one, right? In #338, I rearranged the ordering of `update_all_cells` so that the updating of the `neighbors` lists only needed to happen one time. I did this by moving the `standard_cell_cell_interactions` after the `advance_bundled_phenotype` (and after resolving the birth and death in there).

Though I guess I don't see how that would affect the template project since that should have no interactions...

rheiland:
It was a comment about *both* #338 and #339 - comparing their sims on the model in question, expecting/hoping to see identical results, but only seeing "close". I think #338 may require a zoom walk-through (or else a good night's sleep for me). Seems like a lot of changes, but maybe some verbal explanation would help.

drbergman:
closing in favor of # 338

PR #340, opened 2024-11-27
disallow random detachment from attack

drbergman:
- do not let random detachment occur when one cell is attacking another
- this could result in the detachment occuring while the attack continues
- an alternative solution is to say that detachment necessarily ends the attack

### DIFF
--- a/core/PhysiCell_standard_models.cpp
+++ b/core/PhysiCell_standard_models.cpp
@@ -1441,6 +1441,8 @@ void dynamic_attachments( Cell* pCell , Phenotype& phenotype, double dt )
     while( done == false && j < pCell->state.neighbors.size() )
     {
         Cell* pTest = pCell->state.neighbors[j]; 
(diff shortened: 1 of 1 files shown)


drbergman:
The following demonstrates why we want this change. An `attacker` (right cell) attacks a `defender` (left cell) but the `defender` has a instant detachment rate (`10.0 / min`) and chemotaxes towards the left boundary. However, the attack has not ended and so the `defender` accumulates damage the entire 120 min simulation.

Remake this project with the template (I did it from current `development` branch) and use the attached `cells.csv` and `PhysiCell_settings.txt` (rename to `PhysiCell_settings.xml`).
https://github.com/user-attachments/assets/9926321a-87f3-4468-9fb6-799c24932038
[cells.csv](https://github.com/user-attachments/files/17927704/cells.csv)
[PhysiCell_settings.txt](https://github.com/user-attachments/files/17927710/PhysiCell_settings.txt)

MathCancer:
Fully agreed.

PR #341, opened 2024-12-02
persist attack through attacker transformation

drbergman:
Rebased onto current `development` and **inverted** relative to this PR's original direction: transformation now *keeps* an in-progress attack running instead of ending it.

### Background

- `pAttackTarget` lives in `phenotype.cell_interactions`, so `phenotype = cd.phenotype` in `convert_to_cell_definition` drops it.
- The other two halves of the attack link — the target's `state.attacked_by` and the shared spring — live in `state`, which transformation leaves alone.

Since #422 landed the symmetric attack link, `convert_to_cell_definition` resolved that mismatch by calling `remove_self_from_attacked()` to end the attack. This PR resolves it the other way: hold `pAttackTarget` across the assignment and put it back. All three parts of the link then stay consistent, and the spring is never touched.

This also makes transformation symmetric — attacks *against* a transforming cell already persisted, while attacks *by* it were ended.

`total_damage_delivered` is carried across too (second commit). It also lives in `phenotype.cell_interactions`, so transformation was resetting the attacker's lifetime damage tally to zero — visibly wrong now that the attack it was counting keeps running afterwards.

### Verification

Built and ran `interaction-sample` with `CD8+ T cell -> neutrophil` transformation enabled and a 30 min attack duration. At each transformation of an attacking cell, all three parts of the link survive:

```
TRANSFORM_WITH_ATTACK now=24 newtype=neutrophil target=bacteria \
  pAttackTarget_kept=1 target_lists_me=1 spring_mine=1 spring_theirs=1
```

`interactions` and `rules_sample` are the only sample configs with both nonzero attack and transformation rates, and neither is SVG-diff-checked in CI, so no snapshot changes are expected.

### Known gap

The same run shows the case worth flagging:

```
newtype=neutrophil attack_rate_vs_target=0 attack_damage_rate=1 attack_duration=0.1
```

The transformed cell keeps attacking a cell type its new definition has a **zero** attack rate against — an attack it could never have started — and with unset attack parameters the defaults (`attack_damage_rate = 1.0`, `attack_duration = 30 min`) let that run on.

Handled separately in #432, which warns about exactly this case without changing the simulation.

### DIFF
--- a/core/PhysiCell_cell.cpp
+++ b/core/PhysiCell_cell.cpp
@@ -1129,6 +1129,14 @@ void Cell::convert_to_cell_definition( Cell_Definition& cd )
 	Geometry cell_geometry = phenotype.geometry;
 	Molecular cell_molecular = phenotype.molecular;
 	Custom_Cell_Data cell_custom_data = custom_data;
+
+	// if this cell is attacking, be safe and stop the attack
+	if ( phenotype.cell_interactions.pAttackTarget != NULL )
+	{
+		detach_cells_as_spring(this, phenotype.cell_interactions.pAttackTarget);
+		// phenotype.cell_interactions.pAttackTarget = NULL; // not needed since phenotype is overwritten below
+	}
+
 	// use the cell defaults; 
 	type = cd.type; 
 	type_name = cd.name; 
--- a/core/PhysiCell_cell.cpp
+++ b/core/PhysiCell_cell.cpp
@@ -1158,12 +1158,11 @@ void Cell::convert_to_cell_definition( Cell_Definition& cd )
 	Molecular cell_molecular = phenotype.molecular;
 	Custom_Cell_Data cell_custom_data = custom_data;
 
-	// pAttackTarget is phenotype state, and the assignment below replaces it, so
-	// end this cell's own attack. attacked_by and the spring are in state, which
-	// transformation preserves, so attacks against this cell continue.
-	remove_self_from_attacked();
-	// should we also remove all attackers?  That would be a change in behavior, so for now we don't.
-	// remove_all_attackers();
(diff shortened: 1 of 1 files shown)


drbergman:
Came across this again and adding some more concrete code details.
1. Attack target is part of the attacking cell's `phenotype`. See [here](https://github.com/MathCancer/PhysiCell/blob/69a23dbe7eaac4f41dc9548ad85072eff7921a03/core/PhysiCell_standard_models.cpp#L1269)
2. On transformation, only a subset of `phenotype` fields are protected from conversion. See [here](https://github.com/MathCancer/PhysiCell/blob/69a23dbe7eaac4f41dc9548ad85072eff7921a03/core/PhysiCell_cell.cpp#L1129-L1143)
3. This does not include `phenotype.cell_interactions` ([again](https://github.com/MathCancer/PhysiCell/blob/69a23dbe7eaac4f41dc9548ad85072eff7921a03/core/PhysiCell_standard_models.cpp#L1269), this is where the `pTarget` info is stored)

Thus, the copied `Cell_Definition& cd` `phenotype` with a `NULL` `pTarget` erases our transformed cell's `pTarget`, but the `this.state` is unchanged, and this is where `spring_attachments` is located.

drbergman:
Can we make attack a conserved vs not property?

drbergman:
Need a new PR where the attack persists through transformation.

drbergman:
PR updated to have the attack persist through the transformation. Also had the `total_damage_delivered` persist since it is the same attack.

I do worry this will change model behavior for users: cell that stopped attacking on transformation by default now need to be told to stop attacking (through changing parameters, not requiring custom code). I drafted a warning in #432 that fires when the new cell type has 0 attack rate for the target cell type.

PR #345, opened 2025-01-03
github workflow ubuntu unit tests adjustment from gcc version 11 to version 12

elmbeech:
github lately updated the latest ubuntu version, which by default runs gcc version 12.
this pull request has the required change that the linux os unit tests pass again.


elmbeech:
merged into pull request https://github.com/MathCancer/PhysiCell/pull/349

PR #346, opened 2025-01-04
enable reload setting xml file.

elmbeech:
pull request which replaces the staled pull requests: https://github.com/MathCancer/PhysiCell/pull/344, https://github.com/MathCancer/PhysiCell/pull/298, https://github.com/MathCancer/PhysiCell/pull/264, https://github.com/MathCancer/PhysiCell/pull/250 and https://github.com/MathCancer/PhysiCell/pull/232.

These changes are the bare minimum needed to reload a setting xml file, to be able to run in the same runtime multiple consecutive episodes of a physicell run.

This pull request is compatible with physicell version 1.14.1 and keeps the changes to the existing code bases as small as possible. 
Default settings were chosen so that the current model not will break through these changes.

The code was extensively tested here: https://github.com/elmbeech/episode

### DIFF
--- a/BioFVM/BioFVM_microenvironment.cpp
+++ b/BioFVM/BioFVM_microenvironment.cpp
@@ -253,7 +253,7 @@ bool Microenvironment::get_substrate_dirichlet_activation( int substrate_index )
 // TODO? fix confusing swapped usage of args
 double Microenvironment::get_substrate_dirichlet_value( int substrate_index, int index )
 { 
-    return dirichlet_value_vectors[index][substrate_index]; 
+	return dirichlet_value_vectors[index][substrate_index];
 }  
 
 // new functions for finer-grained control of Dirichlet conditions -- 1.7.0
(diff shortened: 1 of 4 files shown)


drbergman (approved):
I like the changes made here, making the changes more targeted. I just want to document the data that can be re-read in here. I will ignore units and other miscellaneous items that do not affect simulations.

## PhysiCell_settings
- max time
- time steps
- all save parameters
- substrate plot parameters
- omp num threads
- legacy random on unit sphere
- disable_automated_spring_adhesions
- random seed (if an int is given, then all episodes will restart from that seed)
- domain size and voxel size (if these are changed during a `reload`, then they will not be carried over to BioFVM)

## BioFVM microenvironment
- substrate ics
	- if from file, the file folder/name/type cannot be changed between episodes
	- if from XML, does not re-read the XML; just uses the values read in the first time
- DCs
	- enabled and values reset to what was read in the first time

elmbeech:
thank you!

elmbeech:
merged into pull request https://github.com/MathCancer/PhysiCell/pull/349

PR #349, opened 2025-01-12
episode sample project

elmbeech:
This is the physicell episode sample project, written to run multiple consecutive episodes in a single runtime.
This project is linked to pull request https://github.com/MathCancer/PhysiCell/pull/346 , https://github.com/MathCancer/PhysiCell/pull/277 , and https://github.com/MathCancer/PhysiCell/pull/266 .


sample_projects/episode/config/PhysiCell_settings.xml
```diff
@@ -0,0 +1,637 @@
...
+            </dirichlet_nodes>
+        </options>
+    </microenvironment_setup>
+
+    <cell_definitions>
+        <cell_definition name="celltype_a" ID="0">
+            <phenotype>
+                <cycle code="6" name="Flow cytometry model (separated)">
+                    <phase_durations units="min">
+                        <duration index="0" fixed_duration="false">300.030003</duration>
```
drbergman, line 123: it was my fault the template project ended up with these decimals in the durations here. could you reround these to `[300.0, 480.0, 240.0, 60.0]`?

sample_projects/episode/main.cpp
```diff
@@ -0,0 +1,390 @@
...
+		std::cout << "\nrun episode: " << i_episode << " !" << std::endl;
+
+		// variables
+		char filename[1024];
+		std::ofstream report_file;
+		std::vector<std::string> ( *cell_coloring_function )( Cell* ) = my_coloring_function;  // set a pathology coloring function
+		std::string ( *substrate_coloring_function )( double, double, double ) = paint_by_density_percentage;
+
+		// generate output folder
+		std::string folder = "output00" + std::to_string( i_episode % 4 );
```
drbergman, line 114: Whye `% 4`? Is it coincidence that the choice of loop bounds here will not cause these folder names to repeat? Or is that important? Consider variables to hold the value 4 so you can name it something meaningful for users. Even, `total_episodes`.

sample_projects/episode/main.cpp
```diff
@@ -0,0 +1,390 @@
...
+		//PhysiCell_settings.max_time = 1440; //1440; //10080;
+
+		// OpenMP setup
+		omp_set_num_threads( PhysiCell_settings.omp_num_threads );
+
+		if ( !reload )
+		{
+			// setup microenviroment and mechanics voxel size and match the data structure to BioFVM
+			std::cout << "set densities ..." << std::endl;
+			setup_microenvironment( reload );  // modify this in the custom code
```
drbergman, line 144: What does this comment mean? Is the user expected to modify their custom code?

sample_projects/episode/main.cpp
```diff
@@ -0,0 +1,390 @@
...
+			double mechanics_voxel_size = 30;
+			Cell_Container* cell_container = create_cell_container_for_microenvironment( microenvironment, mechanics_voxel_size );
+
+			// load cell type definition and setup tisse
+			std::cout << "load cell type definition and setup tissue ..." << std::endl;
+			generate_cell_types();  // modify this in the custom code
+			setup_tissue();  // modify this in the custom code
+			reload = true;
```
drbergman, line 152: Add comment here highlighting the importance of this step.

sample_projects/episode/main.cpp
```diff
@@ -0,0 +1,390 @@
...
+			microenvironment.display_information( std::cout );
+			double mechanics_voxel_size = 30;
+			Cell_Container* cell_container = create_cell_container_for_microenvironment( microenvironment, mechanics_voxel_size );
+
+			// load cell type definition and setup tisse
+			std::cout << "relaod cell type definition and setup tissue  ..." << std::endl;
+			display_cell_definitions( std::cout );
+			setup_tissue();  // modify this in the custom code
+
+			// set MultiCellDS save options
```
drbergman, line 186: replace these commented lines with a single line comment explaining (or just stating) why we're not re-setting the save options

sample_projects/episode/main.cpp
```diff
@@ -0,0 +1,390 @@
...
+			{
+
+				// max time reached?
+				if ( PhysiCell_globals.current_time > PhysiCell_settings.max_time )
+				{
+					step = false;
+				}
+
+				// on custom time step
+				if ( custom_countdown < diffusion_dt / 3 )
```
drbergman, line 271: making all these triggers `< 0.5 * diffusion_dt` will match with the recent changes throughout the codebase for event triggers

sample_projects/episode/main.cpp
```diff
@@ -0,0 +1,390 @@
...
+				{
+					custom_countdown += custom_dt;
```
drbergman, line 273: Are these countdowns a main addition/feature of this project? Are they essential to the episodic nature of this program? I think these are cool ideas that could be quite useful. I'm just wondering if they're orthogonal to the episodes or not.

drbergman (changes requested):
I'm glad that we're starting to see some of what you're planning with these additions! As this PR requires some of the others that are not yet merged, can you merge them into your `elmbeech:patch1141_episode` so that we can test this project?

Since this project's name does not immediately suggest what it is modeling, could you add a `sample_projects/episode/README.md` to describe what users should be expecting from this? I am not sure what part of this highlights the power of the episodes. To me, all I see is that this runs the model 4 times. Is there more to this that I'm missing? If that's the power (avoiding the overhead of relaunching the program), could you benchmark and provide a simple bar plot, maybe in that readme? Just running `n` sims with and without episodes and comparing wall time.

elmbeech:
Hi Daniel, 

Sure, I can merge all the needed PR's together. 

You are correct, all this sample project does is running consecutive episodes of a physicell model in one runtime.
This is an essential feature if you want to do learning. You simply cannot learn, if you have to destroy the runtime after each episode to reset to the initial condition, because you will at the same time destroy everything you learned from the episode. 
We are on the way to releasing a PhysiCell project that exactly does this, which simply can be installed under user_projects, but to have it work with the regular PhysiCell releases, we first have to get the changes from this PR into the code base. 
The sample project is there for other developers, who want to do similar implementations and, most importantly, as a unit test, to make sure that this feature persists in future PhysiCell releases.

I can add a README with this explanation.

Elmar

sample_projects/episode/config/PhysiCell_settings.xml
elmbeech: i can re-round! definitely.

sample_projects/episode/main.cpp
elmbeech: this comment is from the regular template project, which the episode project is based on.
if the user wants, he can modify, but he is not expected to do so.
what do you think? should I take this comment out?

drbergman: 🤦 I just completely forgot that was there! Nothing to do here.

sample_projects/episode/main.cpp
elmbeech: ok. I can add a comment!

additionally, about the "// modify this in the custom code", the reason is the same as mentioned above.
I will flow the guidance you give above.

sample_projects/episode/main.cpp
elmbeech: ok!

sample_projects/episode/main.cpp
elmbeech: ok!

sample_projects/episode/main.cpp
drbergman: right, and just to be clear (the gh-added context makes it a bit confusing), this was specifically about `reload = true;`

sample_projects/episode/main.cpp
elmbeech: yes, they are essential for the user_project we release soon!
they are not orthogonal to the episodes.
since this sample_project will figure as a template for similar user_projects I would like to leave this in this sample_ project implementation.

drbergman: ok cool! I guess then the to-be-released user_project / sample_project will make it clear how you envision these being used. I don't think you need to modify them here

sample_projects/episode/main.cpp
elmbeech: no worries!

elmbeech:
looks like windows cases some troubles. 
I will look into it.

elmbeech:
merged in pull requests: 
+  https://github.com/MathCancer/PhysiCell/pull/346
+ https://github.com/MathCancer/PhysiCell/pull/345.

elmbeech:
ok @drbergman, now everything is merged, modified as requested, and all tests are passing!

sample_projects/episode/main.cpp
```diff
@@ -115,7 +115,11 @@ int main( int argc, char* argv[] )
 
 		// generate output folder
 		std::string folder = "output" + std::to_string( i_episode );
+		#if defined(_WIN32)
```
drbergman, line 118: we recently added 
```
#if defined(__MINGW32__) || defined(__MINGW64__)
```
for checking OS when creating output directories. Do you have thoughts on if we should also check against this macro, too? Would adding the two above to L118 make your check more robust? I genuinely don't know and don't have (easy) access to a Windows machine to test.

drbergman:
Final question (no promises though 🙃): Does this sample project show the full use of the proposed updates to `initialize_microenvironment` and `load_PhysiCell_config_file`? In other words, will the reinforcement learning project rely on additional calls/functionality of those functions beyond what is in `sample_projects/episode/main.cpp`?

sample_projects/episode/Makefile
```diff
@@ -0,0 +1,325 @@
...
+	rm ALL_CITATIONS.txt 
+	cp ./config/PhysiCell_settings-backup.xml ./config/PhysiCell_settings.xml 
+	touch ./config/empty.csv
+	rm -f ./config/*.csv
+	
+clean:
+	rm -f *.o
+	rm -f $(PROGRAM_NAME)*
+	
+data-cleanup:
```
drbergman, line 203: modify to cleanup the data created by this `output0`, `output1`, etc

sample_projects/episode/Makefile
```diff
@@ -0,0 +1,325 @@
...
+	unzip latest.zip 
+	
+untar: 
+	cp ./archives/latest.tar .
+	tar -xzf latest.tar
+
+# easier animation 
+
+FRAMERATE := 24
+OUTPUT := output
```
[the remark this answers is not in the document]
elmbeech: please take `output3`.
this is the one i specified in the test.yml
then we can be sure that all 4 consecutive episodes were run.

rheiland:
I'd probably prefer to stick with just the `#if defined(_WIN32)` and not use the MINGW-related directives. I can imagine allowing other C++ compilers on Windows someday (I've done past tests using cmake and native Windows C++).

I also confirmed that this sample project compiles, runs, and generates the `output<episode#>` directories on my Windows machine (in Command Prompt and Powershell). But I did not check that the output was "valid". One nitpick might be to consider using leading 0s in your `output` directory filenames (`output_00001`, etc) for easier post-processing.

Also, when running, I see the WARNING related to `<random_seed>` on episodes >= 1 which may be concerning to users, e.g., 
```
run episode: 1 !
(re)set global variables ...
load setting xml config/PhysiCell_settings.xml ...
Using config file config/PhysiCell_settings.xml ...
WARNING: Setting the random seed again.
        You probably have set a user parameter called random_seed.
        Here, we will use the random seed set in user parameters.
        HOWEVER, as of PhysiCell 1.14.0, you should set the random seed in the <options><random_seed> element in the config file.
        Future versions of PhysiCell may throw an error here. Kindly remove the user parameter and just use the <options><random_seed> element.
```
and related to that, do you really want to use the same `random_seed` for every episode and do you want to use just 1 core (`omp_num_threads`)?

elmbeech:
> Final question (no promises though 🙃): Does this sample project show the full use of the proposed updates to `initialize_microenvironment` and `load_PhysiCell_config_file`? In other words, will the reinforcement learning project rely on additional calls/functionality of those functions beyond what is in `sample_projects/episode/main.cpp`?

hmm ...
the main.cpp shows the full use of the proposed update to  `initialize_microenvironment` and `load_PhysiCell_config_file` (and `reset_max_basic_agent_ID`  and `load_PhysiCell_config_file` from earlier already accepted pull requests).
but the reinforcement learning will not run with this main.cpp. it is a bit more complicated what we had to implement for making reinforcement leading possible.
the episode sample project can be regarded as a totally boiled down version of what is absolutely needed to make (reinforcement) learning in physicell possible at all.

is this answering your question?

Makefile
```diff
@@ -225,6 +225,14 @@ immune-function-sample:
 	cp ./config/PhysiCell_settings.xml ./config/PhysiCell_settings-backup.xml 
 	cp -r ./sample_projects/immune_function/config/* ./config/
 
+episode-sample:
```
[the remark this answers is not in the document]
elmbeech: i like singular!

sample_projects/episode/main.cpp
elmbeech: according to what I read, `_WIN32` should work for any windows version (32 and 64 bits) as compiler directive.
but i  am a linux person. i just read up on this. it is a long time ago since i truly installed and run windows. and it is the first time that i have to write C++ software that have to work on windows, linux, and bsd.
so frankly i don't know what is best.

sample_projects/episode/Makefile
elmbeech: ok! i will do.

elmbeech:
> I'd probably prefer to stick with just the `#if defined(_WIN32)` and not use the MINGW-related directives. I can imagine allowing other C++ compilers on Windows someday (I've done past tests using cmake and native Windows C++).

cool! same conclusion on my side.

> 
> I also confirmed that this sample project compiles, runs, and generates the `output<episode#>` directories on my Windows machine (in Command Prompt and Powershell). But I did not check that the output was "valid". One nitpick might be to consider using leading 0s in your `output` directory filenames (`output_00001`, etc) for easier post-processing.
> 

the leading 0 would be nice. could you please give me a hint, how to do this easy in C++?
I have a python scriopt to check if the output is correct (exactely the same when running with one thred). it uses pcdl and the output is at this moment a bit cryptic. i tested heavily on linux. and trust out is the same on windws. this pcdl script will become the unit for the reinforcement learning project and run on github workflow. so, i am not too worried that we not will catch such an error. 

> Also, when running, I see the WARNING related to `<random_seed>` on episodes >= 1 which may be concerning to users, e.g.,
> 
> ```
> run episode: 1 !
> (re)set global variables ...
> load setting xml config/PhysiCell_settings.xml ...
> Using config file config/PhysiCell_settings.xml ...
> WARNING: Setting the random seed again.
>         You probably have set a user parameter called random_seed.
>         Here, we will use the random seed set in user parameters.
>         HOWEVER, as of PhysiCell 1.14.0, you should set the random seed in the <options><random_seed> element in the config file.
>         Future versions of PhysiCell may throw an error here. Kindly remove the user parameter and just use the <options><random_seed> element.
> ```
> 
> (to be clear, there is no random_seed in user params; the warning is printed due to the logic in the C++ code)
> 
I am aware if this waring, but this not really in my hand.

> and related to that, do you really want to use the same `random_seed` for every episode and do you want to use just 1 core (`omp_num_threads`)?

yes, if you wanna check that you have no drifet in concesutive episodes you wanna run all episodes with the same seed and one thred. 
and if you wann do learning you will have to run totaly random anyway. so I think this is ok.
also, it is a sample project. if someone wants to change this behavior, he can simply modify it.

drbergman:
> hmm ... the main.cpp shows the full use of the proposed update to `initialize_microenvironment` and `load_PhysiCell_config_file` (and `reset_max_basic_agent_ID` and `load_PhysiCell_config_file` from earlier already accepted pull requests).

In this case, we can choose to refactor these functions rather than redefine them. This avoids introducing in these functions what I still find to be confusing delineations between what is and is not reloadable. See [#346](https://github.com/MathCancer/PhysiCell/pull/346#pullrequestreview-2531265026) for what is reloadable. My first pass replaces parts of your `main.cpp` with 
```
if ( i_episode == 0 )
{
	XML_status = load_PhysiCell_config_file( settingxml );
}
else
{
	XML_status = read_PhysiCell_config_file( settingxml ); // NEWLY DEFINED (SEE BELOW)
	if (XML_status)
	{
		PhysiCell_settings.read_from_pugixml();
		create_output_directory( PhysiCell_settings.folder ); // could omit since elsewhere in main.cpp the output folder is created (or could omit that line instead)
	}
}
```
```
// reset microenvironment and mechanics voxel size and match the data structure to BioFVM
std::cout << "reset densities ..." << std::endl;
set_microenvironment_initial_condition(); // NEWLY DEFINED (does everything in set_microenvironment below the `if (!reload)` block)
microenvironment.display_information( std::cout );
```
where we define
```
bool read_PhysiCell_config_file( std::string filename )
{
	physicell_config_dom_initialized = false; 

	std::cout << "Using config file " << filename << " ... " << std::endl ; 
	pugi::xml_parse_result result = physicell_config_doc.load_file( filename.c_str()  );
	
	if( result.status != pugi::xml_parse_status::status_ok )
	{
		std::cout << "Error loading " << filename << "!" << std::endl; 
		return false;
	}
	
	physicell_config_root = physicell_config_doc.child("PhysiCell_settings");
	physicell_config_dom_initialized = true; 
	return true
}
```
I opened a [PR](https://github.com/elmbeech/PhysiCell/pull/1) showing the full set of changes for this refactor.

drbergman:
> Also, when running, I see the WARNING related to `<random_seed>` on episodes >= 1 which may be concerning to users, e.g.,

It only shows up on `i_episode==1` for me. the `SeedRandom` has a `static bool warned` to make sure the user only gets warned once. If we want to avoid even that, we could move that (or probably the other `static bool setup_done` variable) into global scope and write a function to toggle its state.

.github/workflows/tests.yml
```diff
@@ -34,6 +34,7 @@ jobs:
           {project: "pred-prey-farmer", name: "PhysiCell prey predator", binary: "pred_prey", config: "config/PhysiCell_settings.xml", max_time: 120, output_folder: ""},
           {project: "rules-sample", name: "PhysiCell rules sample", binary: "rules_sample", config: "config/PhysiCell_settings.xml", max_time: 120, output_folder: ""},
           {project: "asymmetric-division-sample", name: "PhysiCell asymmetric division", binary: "project", config: "config/PhysiCell_settings.xml", max_time: 120, output_folder: ""}
```
drbergman, line 36: add `,` at end

.github/workflows/tests.yml
```diff
@@ -34,6 +34,7 @@ jobs:
...
+          {project: "episode", name: "PhysiCell Episode", binary: "project", max_time: 120, config: "config/PhysiCell_settings.xml", output_folder: "output3"},
```
drbergman, line 37: `project: "episode-sample"`

drbergman: also, including the output folder here means you want to compare the gh-generated svgs against those in `tests/cases/output_episode-sample/`. Either add those svgs or set `output_folder: ""` as others above do

drbergman:
still need to either include test svgs in `tests/cases/output_episode-sample/` or set the `output_folder: ""` in `tests.yml`

sample_projects/episode/main.cpp
```diff
@@ -0,0 +1,394 @@
...
+		if ( i_episode == 0 )
+		{
+			// load xml file
+			std::cout << "load setting xml " << settingxml << " ..." << std::endl;
+			bool XML_status = false;
+			XML_status = load_PhysiCell_config_file( settingxml );
+			PhysiCell_settings.folder = folder;
+			create_output_directory( PhysiCell_settings.folder );
+
+`			// OpenMP setup
```
drbergman, line 137: extra backtick here. sorry if it was from me!

elmbeech: yes, was not aware of the `create_output_directory` function?
i don't understand the "extra bracketing". looks all good to me.

elmbeech:
> still need to either include test svgs in `tests/cases/output_episode-sample/` or set the `output_folder: ""` in `tests.yml`

I thought had already updated the yml files.
please check for folder `output3` then we are sure that really consecutive episodes were run!

sample_projects/episode/main.cpp
drbergman: L137 has an extra backtick at the front. you fixed it

drbergman:
> please check for folder `output3` then we are sure that really consecutive episodes were run!

That works manually, but the gh actions will want to test the svgs in `output3` against what is in `tests/cases/output_episode-sample/` because of the line in the `tests.yml` matrix for `episode-sample`. See the failed actions.

sample_projects/episode/Makefile
elmbeech: @drbergman  i would not change this!
it is dangerous to write `rm -rf ./output*` or something like this!
usually, when people rename the output folder they don't wanna have it deleted by `make data-clean`.

.github/workflows/tests.yml
elmbeech: @drbergman is this already fixed?

sample_projects/episode/Makefile
elmbeech: as discussed!

elmbeech:
this code passes all the tests.
i think we are pretty much there. 
thank you!

drbergman (approved):
Looks good to me now. Refactors in the code improve code readability and make the necessary internals exposed for this project. A sample project that gives the framework a skeleton upon which to build will make this easier for users to pick up. Also, put the new project in the testing suite for CI.

elmbeech:
Thank you, Daniel!

MathCancer:
NICE WORK!!

elmbeech:
Thank you, Paul!

elmbeech:
I could successfully compile the reinforcement learning user_project.
The only thing I cannot make compatible (without going here into the details) with the current episode sample project is: 
```
std::string settingxml  # C++
```
but to be compatible will have to make it in C style like 
```
char *settingxml
```

MathCancer:
Elmar thank you for the OUTSTANDING work on this, as well as for being so responsive on edits. 

Daniel, thank you for all the mentoring and QC!!!

PR #350, opened 2025-01-14
fix event trigger in asym div ex

drbergman:
use new 1.14.1 standard save event triggers in the asymmetric division sample project.


MathCancer:
Please note that this *may* have some issues with PhysiBoSS, since those automated checks did not complete.

PR #351, opened 2025-01-14
re-round template project cycle durations

drbergman:
for a brief second, studio was parsing cycle durations into rates and back, truncating to 5 decimals at each step. studio no longer double-converts cycle durations.
the template project was updated to 1.14.1 in this window and the cycle durations go weird decimals. This restores them


rheiland:
Good catch! PR looks good and should hopefully pass tests once https://github.com/MathCancer/PhysiCell/pull/345 is merged.

PR #352, opened 2025-01-15
more robust macro to check for windows

drbergman:
When creating output directories, currently use `__MINGW32__` and `__MINGW64__` to check if compiling on Windows, but this only checks if using particular windows compilers. they are the ones we tell users to use, but switching to `_W32` will be more robust.


rheiland:
I agree. And I tested your PR on Windows - it works.

PR #357, opened 2025-02-02
DC initialization from file

drbergman:
provide a CSV file with similar structure to the substrate ICs csv header row: x,y,z,<substrate_01>,<substrate_02>,... where the list of substrates need not be all the substrates, only those with DCs being set. Each subsequent row is:
```
<x_coord>,<y_coord>,<z_coord>,<val_01>,<val_02>,...
```
where the coords must be specified and the vals can either be empty (<val_01>,,<val_03>,...) or a number
- If empty, then nothing chagnes for the substrate in that column at that voxel
- If a number, then the DC activation for that voxel-substrate pairing is set to the number

If the XML sets DCs, this will overwrite those values **for which a number is supplied in the CSV**. In other words, if a voxel on the boundary has a DC set to ON by the XML, parsing the CSV file can only change the value of the DC, not turn it OFF.

a sample project in `sample_projects/dirichlet_from_file` is provided. try it with `make dirichlet-from-file-sample`


BioFVM/BioFVM_microenvironment.cpp
```diff
@@ -1225,6 +1231,195 @@ void set_microenvironment_initial_condition( void )
...
 	}
 
+	// set the dirichlet condition
+	set_dirichlet_initial_condition();
+
+	microenvironment.display_information(std::cout);
+	return;
+}
+
+void load_initial_conditions_from_matlab(std::string filename)
```
drbergman, line 1241: This function is unchanged except that the error at the bottom now goes to `std::cerr` rather than `std::cout`. It is green here only because the `set_microenvironment_initial_condition` function is refactored and that moved things around too much for git to track well.

BioFVM/BioFVM_microenvironment.cpp
```diff
@@ -1225,6 +1231,195 @@ void set_microenvironment_initial_condition( void )
...
+				exit(-1);
+			}
+		}
+		voxel_set.push_back(voxel_ind);
+	}
+
+	return;
+}
+
+void load_initial_conditions_from_csv(std::string filename)
```
drbergman, line 1292: This function is unchanged except the comment `// Do not include a header row.` has been removed since you can (and probably should) included a header row. See comment on [`load_initial_conditions_from_matlab`](https://github.com/MathCancer/PhysiCell/pull/357/files#r2102389603).

Note: you should include a header row to explicitly declare the substrates you are setting, rather than relying on their index in the substrate list.

BioFVM/BioFVM_microenvironment.cpp
```diff
@@ -1225,6 +1231,195 @@ void set_microenvironment_initial_condition( void )
...
+				  << "\tRemember, your table should have dimensions #voxels x (3 + #densities)." << std::endl;
+		exit(-1);
+	}
+
+	file.close(); 	
+	
+	return;
+}
+
+void get_row_from_substrate_initial_condition_csv(std::vector<int> &voxel_set, const std::string line, const std::vector<int> substrate_indices, const bool header_provided)
```
drbergman, line 1379: This function is entirely unchanged. See comment on [`load_initial_conditions_from_matlab`](https://github.com/MathCancer/PhysiCell/pull/357/files#r2102389603).

BioFVM/BioFVM_microenvironment.cpp
```diff
@@ -1225,6 +1231,195 @@ void set_microenvironment_initial_condition( void )
...
+	voxel_set.push_back(voxel_ind);
+}
+
+void set_dirichlet_initial_condition( void )
+{
+	set_dirichlet_boundaries_from_XML();
+	set_dirichlet_boundaries_from_file();
+}
+
+void set_dirichlet_boundaries_from_XML( void )
```
drbergman, line 1419: This new function is just the old end to `set_microenvironment_initial_condition`, refactored into its own function for simplicity and clarity of logic. This is especially useful as its caller (`set_dirichlet_initial_condition`) handles the loading of DC conditions from both XML and file.

BioFVM/BioFVM_microenvironment.cpp
```diff
@@ -1391,70 +1586,48 @@ void set_microenvironment_initial_condition( void )
...
 		}
-		
 	}
-
-	microenvironment.display_information(std::cout);
-	return;
 }
 
-void load_initial_conditions_from_matlab(std::string filename)
+void set_dirichlet_boundaries_from_file( void )
```
drbergman, line 1590: See comment on [`load_initial_conditions_from_matlab`](https://github.com/MathCancer/PhysiCell/pull/357/files#r2102389603). This is a new function, not a rework of `load_initial_conditions_from_matlab`. It just shares enough code that git thinks it is reworking this.

BioFVM/BioFVM_microenvironment.cpp
drbergman: In reviewing this PR, I found an error when loading DCs from a CSV without a header row. That mistake was traced back to the code for reading in substrates from a CSV (written by yours truly). So, that has been fixed here.

BioFVM/BioFVM_microenvironment.cpp
drbergman: need to fix this. should be
```
<< "\tExpected: " << microenvironment.number_of_densities() << std::endl
```

drbergman:
After merging this and #386, there is some opportunity to reduce the codebase through DRY principles. Not worth stacking them here, so will just wait on that.

PR #359, opened 2025-02-03
msys2 package list clean out and brew gcc update to v14.

elmbeech:
lately i was writing new physicell installation manuals for the power users.
thereby, i cleaned out the msys2 package list.
additionally, i noticed that brew nowadays ships with gcc version 14.
this is an update to the workflow files from what i learned on the way.

### DIFF
--- a/.github/workflows/build_binaries.yml
+++ b/.github/workflows/build_binaries.yml
@@ -82,7 +82,7 @@ jobs:
     - uses: msys2/setup-msys2@v2
       with:
         update: true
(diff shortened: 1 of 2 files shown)


drbergman (approved):
This gets my approval. It's just changing the download packages for a windows install to run the test suite. If anyone with windows experience objects, please feel free to raise.

elmbeech:
i just checked. brew nowadays ships with gcc version 16.
so maybe this could be updated (gcc@16 instead of gcc@14 as seems to be nowadays).

+ https://formulae.brew.sh/formula/gcc

the msys2 chaneges are still up to date.

PR #361, opened 2025-02-11
fix ode-energy-sample and clean/fix librr stuff

drbergman:
Update [25-02-13]: See PR #362 for a better update to the libRR api.

- introduce `update_time_step` in `RoadRunnerIntracellular`
    - set for each instance of `RoadRunnerIntracellular`
    - this controls the timestep for the model
    - will be set to `intracellular_dt` if not specified in the xml for the given cell type
    - add the `intracellular_dt` element as a child of the `intracellular` element in the xml
    - could be set on a cell-by-cell basis if desired
- introduce `previous_update_time` in `RoadRunnerIntracellular`
    - tracks the previous update time
    - used to determine how long to run the ODE when updating
    - set to `current_time` when copying/cloning
- fix `next_librr_run` in `RoadRunnerIntracellular`
    - was not being updated and thus ODE models were running at every time step
    - check `current_time >= next_librr_run - 0.5 * diffusion_dt` to determine if an update is needed
    - set to `current_time + update_time_step` when copying/cloning
- cleaned up librr files
- fixed and cleand ode-energy-sample
    - it was running the intracellular models in `main.cpp` and in the update cells function
    - it seemed to try to set intracellular_dt inside the intracellular element of the xml
    - messy custom.cpp and custom.h

### DIFF
--- a/addons/libRoadrunner/src/librr_intracellular.cpp
+++ b/addons/libRoadrunner/src/librr_intracellular.cpp
@@ -8,58 +8,43 @@ RoadRunnerIntracellular::RoadRunnerIntracellular() : Intracellular()
 	intracellular_type = "sbml";
     std::cout << "====== " << __FUNCTION__ << "() intracellular_type=" << intracellular_type << std::endl;
     std::cout << "====== " << __FUNCTION__ << "() sbml_filename = " <<  sbml_filename << std::endl;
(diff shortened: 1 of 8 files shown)


drbergman:
See PR #362 for a better update to the libRR api.

PR #362, opened 2025-02-13
Improve librr api

drbergman:
The key improvement here is the **automated** linking of PhysiCell values and SBML values through the config file (NO CUSTOM CODING REQUIRED!!). It does introduce a breaking change in that the attribute names are now uniformly `physicell_name` (in place of `PC_<key>`) and a `type` attribute is also required.

To update previous config files, users must change the old tokens attribute names (`PC_<key>` format). The old tokens still work. In addition, the `type` attribute must be included and set to one of `io`, `input`, or `output`. This last step is what automatically allows the code to pass values between PhysiCell and SBML models without further user input. Because of that, users would also likely need to (get to) delete the custom code that passed the values back and forth.

Note: users would still need to ensure the `Makefile` has the libRoadRunner macro, compile target, and other updates (see the `sample_projects_intracellular/ode/ode_energy/Makefile` for example)

### Updated xml
Previously, a typical child element of the `<intracellular>` element would look like
```
<map PC_<key>="oxygen" sbml_species="Oxygen"></map >
```
where `<key>` could be any of `substrate`, `phenotype`, or `custom_data`.
Now, that element would be
```
<map type="io" physicell_name="intracellular oxygen" sbml_species="Oxygen"></map>
```
Some notes:
- `type` is any of `input`, `output`, or `io` and defines whether the mapping is applied before or after updating the intracellular model AND whether PhysiCell values are being _input_ into the model or _output_ from the model (or both in the case of `io`)
- `physicell_name`
  - matches the convention in the MaBoSS addon
  - can be any **signal** OR **behavior** dictionary entry
    - only a subset of signals are available to be assigned in an output mapping
    - those available for output:
      - `intracellular <substrate_name>` (assume sbml species is concentration and sets the total internalized using current cell volume)
      - `volume` (rescales all volumes)
      - `damage`
  - tokens from the previous version are also accepted, but only the volume ones (`vtsc`, `vtsn`, and `vff`) should be considered best practice
    - streamlines validation of the models to ensure no two mappings set the same value
    - the other tokens could be replaced by signals/behaviors.
  - among all output mappings (`output` or `io`), the `physicell_name`'s must be unique
- `sbml_species`
  - same convention as before
  - among all input mappings (`input` or `io`), the `sbml_species`'s must be unique

### DIFF
--- a/addons/libRoadrunner/src/librr_intracellular.cpp
+++ b/addons/libRoadrunner/src/librr_intracellular.cpp
@@ -569,17 +569,18 @@ void RoadRunnerIntracellular::initialize_intracellular_from_pugixml(pugi::xml_no
 
         node_map = node_map.next_sibling("map");
     }
-    validate_mappings(new_input_mappings);
-    validate_mappings(new_output_mappings);
+    validate_mappings(new_input_mappings, true);
+    validate_mappings(new_output_mappings, false);
 
     input_mappings = std::move(new_input_mappings);
     output_mappings = std::move(new_output_mappings);
 }
 
-void validate_mappings(std::vector<RoadRunnerMapping*> mappings)
+void validate_mappings(std::vector<RoadRunnerMapping *> mappings, bool is_inputs)
 {
+    if (mappings.empty())
+    { return; }
     std::vector<std::string> values_already_set;
-    bool is_inputs = mappings[0]->io_type=="input";
     std::string name_to_add;
     for (auto mapping : mappings)
     {
--- a/addons/libRoadrunner/src/librr_intracellular.cpp
+++ b/addons/libRoadrunner/src/librr_intracellular.cpp
@@ -72,28 +72,24 @@ void RoadRunnerMapping::initialize_mapping( void )
 // PhysiCell does not have an API for setting signals, but libRoadRunner does in limited cases
 MappingFunction select_signal_setter(const std::string& name, const std::string& sbml_species)
 {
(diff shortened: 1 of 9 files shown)


rheiland:
Still reviewing, but really liking this so far! Obviously, we'll need to synchronize changes for the Studio at release. I do wonder if we should expand the scope of this sample project a bit in order to demo/test for other options, e.g., `<map type="input" ... `. 

Unless I'm missing something, I don't think we want `dt_intracellular` in the `<overall>`:
https://github.com/drbergman/PhysiCell/blob/improve-librr-api/sample_projects_intracellular/ode/ode_energy/config/PhysiCell_settings.xml#L94

drbergman:
Great! And roger that on the studio updates. Hopefully pretty straightforward since we're now using signals/rules.

I think that's a good idea to at least have one `<map type="input"` element. However, I'm failing to think of a reasonable input for this model. We want it to stay biologically accurate...

On `intracellular_dt`, the element in the `<overall>` node sets the `PhysiCell::intracellular_dt` variable, which now that I'm looking at that more closely did nothing until this PR uses it as the default libRR `update_time_step` (for when no `<intracellular_dt>` node was found in `<intracellular>`). We could enforce explicit setting of this for each cell type rather than falling back on `PhysiCell::intracellular_dt`.

drbergman:
Will merge once conflicts resolved.

drbergman:
will merge when the checks pass

PR #364, opened 2025-02-18
update intracellular in first time step

drbergman:
Low priority PR that fixes bug causing intracellular updates to skip the first diffusion time step at t=0. I don't believe that PhysiBoSS or PhysiCelldFBA require this step to be skipped, but please inform if it is necessary :) 

It is possible that rules that depend on variables expected to be set in a `post_intracellular_update` function could receive unexpected inputs causing unintended behavior in the first `advance_bundled_phenotype_functions` and subsequent mechanics updates.

- this change allows intracellular updates to occur in the first diffusion step
- before, they had to wait one diffusion step to update (if there's a reason for this, it's lost on me)
- removed need for `initialzed` [sic] field in `Cell_Container` class
- also make easier-to-parse logic and variable names for when to update phenotype and mechanics

### DIFF
--- a/core/PhysiCell_cell_container.cpp
+++ b/core/PhysiCell_cell_container.cpp
@@ -124,6 +124,14 @@ void Cell_Container::update_all_cells(double t, double phenotype_dt_ , double me
 {
 	// secretions and uptakes. Syncing with BioFVM is automated. 
 
+	static double time_since_last_phenotype = phenotype_dt_;
+	static double time_since_last_mechanics = mechanics_dt_;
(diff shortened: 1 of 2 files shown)


vincent-noel:
I don't think there is a reason, and I don't see how.
In the case of PhysiBoSS, if there is a need to wait, there is a special parameter which can be set (start_time) for example to allow for a gradient to reach equilibrium.

drbergman:
Merge once conflict is resolved.

drbergman:
waiting for checks to past before merging. had to bring in the change to pre-update intracellular first

Issue #214, opened 2024-02-26
DC function standardization


drbergman:
There are many functions provided in `BioFVM_microenvironment.cpp` designed for implementing Dirichlet conditions (DCs). However, there is inconsistency across what they do and the naming conventions shed little light on what their under-the-hood differences are. Case in point (and maybe even the whole enchilada):
* `add_dirichlet_node` does not update `dirichlet_activation_vectors` and thus **cannot** be used to start applying a DC
* `update_dirichlet_node( int voxel_index , int substrate_index , double new_value )` does do this but...
* `update_dirichlet_node( int voxel_index , std::vector<double>& new_value )` does not

rheiland:
This goes back to https://github.com/MathCancer/PhysiCell/issues/124 and my refactor of `BioFVM_microenvironment.{h/cpp}` to fix the DC bug that you had reported then. However, my refactoring did not include making any changes to the two flavors of `update_dirichlet_node` (one for a scalar and one for a vector), i.e., this from 1.11.0 is the same as the current 1.13.1 version:
```
void Microenvironment::update_dirichlet_node( int voxel_index , std::vector<double>& new_value )
{
	mesh.voxels[voxel_index].is_Dirichlet = true; 
	dirichlet_value_vectors[voxel_index] = new_value; 
	
	return; 
}

void Microenvironment::update_dirichlet_node( int voxel_index , int substrate_index , double new_value )
{
	mesh.voxels[voxel_index].is_Dirichlet = true; 
	dirichlet_value_vectors[voxel_index][substrate_index] = new_value; 
	
	dirichlet_activation_vectors[voxel_index][substrate_index] = true; 
	
	return; 
}
```
We need to clearly define and document what updating an entire vector of DCs even means. Does "update" imply they will also always be activated, i.e., set `true`,  or does it mean the *values* will be updated but not necessarily their activation? I remember spending a LOT of time on the bug fix earlier and just want to make sure we're in agreement on what to do now.

Fwiw, here's a comment from the User Guide (which is somewhat deprecated of course):
```
Note that turning on or off a Dirichlet condition for a substrate applies it at all Dirichlet nodes for which
is_dirichlet_node(int voxel_index) is true.
```

PR #369, opened 2025-04-01
simplified dc functions

drbergman:
The current API for the Dirichlet nodes is inconsistent and confusing (See #214). To simplify the API for users, I have created two functions that handle all the DC logic in a straightforward, intuitive, and consistent manner. Several variants are included for those looking to fully optimize these calls (though unless the model is changing DC activations nearly every `diffusion_dt`, I doubt they matter enough to be noticed). The old API, using verbs `add`, `update`, and `remove` remain so that we can preserve backwards compatibility. The new API uses `fix` and `unfix`.

## New API
To fix a substrate at a voxel, i.e., set a DC condition, users can call
```cpp
fix_substrate_at_voxel(std::string substrate_name, int voxel_index[, double new_value]);
```
where `new_value` is optional[^1].

To unfix a substrate at a voxel:
```cpp
unfix_substrate_at_voxel(std::string substrate_name, int voxel_index);
```
In these and all cases that follow, passing in `int substrate_index` in place of `substrate_name` is allowed.

## Handling Boolean activations
The confusing part of the previous API was the inconsistency of updating the Booleans that tracked which `(voxel, substrate)`, `voxel`, and `substrate` had active DCs. This new proposed API handles these consistently:
- `dirichlet_activation_vectors` maintains the lowest level detail where `dirichlet_activation_vectors[voxel_index][substrate_index]` indicates whether that substrate is has a DC at that voxel
- `dirichlet_activation_vector` maintains whether a substrate has a DC anywher and is accessed by `dirichlet_activation_vector[substrate_index]`
- `voxel.is_Dirichlet` maintains whether any substrate has a DC at the given voxel

## Convenience API to (possibly) save time
Maintaining the above Boolean activation values can incur unnecessary computational expense if `fix`-ing or `unfix`-ing many substrates/voxels at once. For users who wish to save even these scraps of time, the following are implemented to update the Boolean variables after accounting for all the changes:
```cpp
fix_substrate_at_voxels(std::string substrate, std::vector<int>& voxel_indices, std::vector<double>& new_values); // see below for alternative signatures
unfix_substrate_at_voxels(std::string substrate, std::vector<int>& voxel_indices);
```

## Fix with flexible value update
Anticipating users turning "back on" a DC, the `fix_...` calls do not require passing in a value. In this case, whatever value was in `dirichlet_value_vectors[voxel_index][substrate_index]` remains.
```cpp
fix_substrate_at_voxel( std::string substrate, int voxel_index );
fix_substrate_at_voxels( std::string substrate, std::vector<int>& voxel_indices );
```
Similarly, if fix multiple voxels at once with uniform DCs, just the one value can be passed:
```cpp
fix_substrate_at_voxels( std::string substrate, std::vector<int>& voxel_indices, double new_value );
```

## All substrates in a voxel at once
To support changing all substrates in a voxel at once, the following three functions are provided:
```cpp
fix_substrates_at_voxel( int voxel_ind, std::vector<double>& new_values );
fix_substrates_at_voxel( int voxel_ind );
unfix_substrates_at_voxel( int voxel_index );
```

[^1]: It does not use an optional arg, but just has two implementations

### DIFF
--- a/BioFVM/BioFVM_microenvironment.cpp
+++ b/BioFVM/BioFVM_microenvironment.cpp
@@ -232,6 +232,11 @@ void Microenvironment::fix_substrate_at_voxel( int substrate_index, int voxel_in
 
 void Microenvironment::fix_substrates_at_voxel( int voxel_index , std::vector<double>& new_values )
 {
+	if (new_values.size() != dirichlet_value_vectors[voxel_index].size())
+	{
+		std::cerr << "Error: Incorrect number of values passed in to fix_substrates_at_voxel. Expected " << dirichlet_value_vectors[voxel_index].size() << " values, but got " << new_values.size() << "." << std::endl;
+		return;
+	}
 	dirichlet_value_vectors[voxel_index] = new_values;
 	return fix_substrates_at_voxel( voxel_index );
 }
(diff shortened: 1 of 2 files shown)


drbergman:
Accepting this should precipitate changes to #357 to make use of this simplified API

PR #373, opened 2025-05-13
Fix attack no detach

drbergman:
In PR #340, the fix targeted the wrong function for disallowing random detachment during an attack. Not sure how I made that mistake, but now it's fixed. As of opening this PR, there are two commits:
- the first addresses the fix
- the second improves the code in `dynamic_attachments` and `dynamic_spring_attachments` so that the code is more straightforward

If the code refactor is not desired, I can roll back that commit easily enough.

To prove this has fixed the issue, here is a movie from the model provided in #340 in which the spring attachment persists throughout the attack duration from the updated code despite the high detachment rate.

https://github.com/user-attachments/assets/3aacf98e-c291-444c-a16c-39cb8ba08f87

### DIFF
--- a/core/PhysiCell_standard_models.cpp
+++ b/core/PhysiCell_standard_models.cpp
@@ -1422,87 +1422,55 @@ void standard_asymmetric_division_function( Cell* pCell_parent, Cell* pCell_daug
 void dynamic_attachments( Cell* pCell , Phenotype& phenotype, double dt )
 {
     // check for detachments 
-    double detachment_probability = phenotype.mechanics.detachment_rate * dt; 
-    for( int j=0; j < pCell->state.attached_cells.size(); j++ )
-    {
-        Cell* pTest = pCell->state.attached_cells[j]; 
-        if( UniformRandom() <= detachment_probability )
-        { detach_cells( pCell , pTest ); }
-    }
+    double detachment_probability = phenotype.mechanics.detachment_rate * dt;
+	for( auto pTest : pCell->state.attached_cells )
+	{
+		if( UniformRandom() <= detachment_probability )
+		{ detach_cells( pCell , pTest ); }
+	}
 
-    // check if I have max number of attachments 
-    if( pCell->state.attached_cells.size() >= phenotype.mechanics.maximum_number_of_attachments )
-    { return; }
-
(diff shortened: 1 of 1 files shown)


core/PhysiCell_standard_models.cpp
```diff
@@ -1422,87 +1422,55 @@ void standard_asymmetric_division_function( Cell* pCell_parent, Cell* pCell_daug
...
-
-    // check for new attachments; 
-    double attachment_probability = phenotype.mechanics.attachment_rate * dt; 
-    bool done = false; 
-    int j = 0; 
-    while( done == false && j < pCell->state.neighbors.size() )
-    {
-        Cell* pTest = pCell->state.neighbors[j]; 
-		if (phenotype.cell_interactions.pAttackTarget==pTest || pTest->phenotype.cell_interactions.pAttackTarget==pCell) // do not let attackers detach randomly
-		{ continue; }
```
drbergman, line 1445: this `if` block is the wrong fix from #340. It is in the `dynamic_attachment` function but attacks result in spring attachments. Furthermore, the logic is applied to prevent `pTest` and `pCell` from attaching again rather than from preventing them from spontaneously detaching.

core/PhysiCell_standard_models.cpp
```diff
@@ -1422,87 +1422,55 @@ void standard_asymmetric_division_function( Cell* pCell_parent, Cell* pCell_daug
...
     // check for detachments 
-    double detachment_probability = phenotype.mechanics.detachment_rate * dt; 
-    for( int j=0; j < pCell->state.spring_attachments.size(); j++ )
-    {
-        Cell* pTest = pCell->state.spring_attachments[j]; 
+    double detachment_probability = phenotype.mechanics.detachment_rate * dt;
+	for( auto pTest : pCell->state.spring_attachments )
+	{
+		if( phenotype.cell_interactions.pAttackTarget==pTest || pTest->phenotype.cell_interactions.pAttackTarget==pCell ) // do not let attackers detach randomly
+		{ continue; }
```
drbergman, line 1455: Here's the corrected check to ensure that spring attachments to not spontaneously detach when they were formed for attacks.

PR #379, opened 2025-07-24
fix parametrization of UniformInShell function

heberlr:
This fix enables sampling cells within a spherical shell defined by an inner radius r₁ and an outer radius r₂.


core/PhysiCell_rules.cpp
```diff
@@ -2331,13 +2331,13 @@ std::vector<double> UniformInShell( double r1, double r2 )
 	sqrt_one_minus_T -= T;
```
drbergman: I'm not sure I understand what the `T` does here. Could we make use of `UniformOnUnitSphere` to simplify this? Basically, multiple the return value of `UniformOnUnitSphere` by your `param1` here?

PR #382, opened 2025-08-19
unused old_position

rheiland:
* in `Cell::update_position`, comment out the unused `std::vector<double> old_position(position)` to avoid confusion and perhaps speed up a simulation


drbergman (approved):
Makes complete sense to remove this. I second that this is unnecessary.

I personally would prefer to delete this line and the three commented out lines below. It seems they offer an explanation for this variable: that it was created for debugging/diagnostic purposes.

PR #383, opened 2025-08-25
add cell_division_direction_function for divide()

rheiland:
* add a cell custom function as an alternative to randomly positioning the daughter cell in  `divide()`
* if this function is defined (non-NULL), it will be called and return a 3-vector which is the normalized direction from the parent cell to the daughter cell
* this PR does not define another sample project to test/demo; rather, I'd like to suggest that we consider adopting a .zip approach that bundles all necessary files for compiling and testing. For now, we could simply adopt the `user_projects` workflow.

To test:
* clone this branch, then:
```
make
make reset
# download https://github.com/rheiland/sample_models/blob/main/bacterial_growth.zip
# and put into `/user_projects` and  unzip it there, then:
make load PROJ=bacterial_growth
make
project    # or use the Studio to run & plot results
```
Sample output:
Screenshot 2025-08-28 at 11 22 26 AM


core/PhysiCell_cell.cpp
```diff
@@ -608,10 +608,19 @@ Cell* Cell::divide( )
...
-		rand_vec[1]*state.orientation[1]+rand_vec[2]*state.orientation[2])*state.orientation;	
+
+	std::vector<double> rand_vec; 
+	if( this->functions.cell_division_direction_function )
+	{ 
+		rand_vec = this->functions.cell_division_direction_function( this ); 
+	}
+	else
+	{
+		rand_vec = cell_division_orientation();
```
drbergman, line 622: any reason not to initialize `cell_division_direction_function` as `cell_division_orientation`? simplifies the logic here? I know we follow this structure you have here elsewhere in the codebase, but I think setting the default and overriding is simpler and yields simpler logic here.

core/PhysiCell_cell.cpp
```diff
@@ -608,10 +608,19 @@ Cell* Cell::divide( )
...
 	*/
-	
-	std::vector<double> rand_vec = cell_division_orientation(); 
-	rand_vec = rand_vec- phenotype.geometry.polarity*(rand_vec[0]*state.orientation[0]+ 
-		rand_vec[1]*state.orientation[1]+rand_vec[2]*state.orientation[2])*state.orientation;	
+
+	std::vector<double> rand_vec; 
+	if( this->functions.cell_division_direction_function )
+	{ 
+		rand_vec = this->functions.cell_division_direction_function( this );
```
drbergman, line 615: do we trust the user to return a unit vector here? the logic with the radius below assumes this is a unit vector

drbergman (changes requested):
This is actually quite timely and really useful! I think a couple minor changes will improve on this in ways worth considering though. Looking forward to discussing further!

rheiland:
Thanks for the feedback! Offhand, I'm OK with both of your suggestions: init `cell_division_direction_function` as`cell_division_orientation`, and normalizing the vector returned by the custom function (even if they already did) - a safer approach. I'll sit on it awhile in case others chime in. I'm also curious to hear if I've missed something more fundamental, as this topic came up in a Macklin lab mtg recently.

drbergman:
I'm curious to hear as well...the only other thing that comes to mind is managing the final `rand_vec` orientation in a 2D sim to be in the xy-plane.

rheiland:
Actually, that might be one argument for keeping the if-else logic on the `cell_division_direction_function` in `divide()`: we could explicitly check the user-supplied vector for being in the x-y plane for a 2D sim (which implies additional checks for 2D/3D sim and z ~= 0.0). Ugh.

drbergman:
we could write a setter function that sends a default cell through the function assert these condition(s) hold. Normalization regardless. in xy-plane when in 2D.

PR #385, opened 2025-09-05
Feature extended asym div

drbergman:
extended asym div that allows for any combo of cell types for **both** daughter cells. Thank you @davidlzhou for contributing!

See the new sample project for how to use.

# Config
```xml
<cycle code="5" name="live">
  <phase_transition_rates units="1/min">
    <rate start_index="0" end_index="0" fixed_duration="true">0.005</rate>
  </phase_transition_rates>
  <extended_asymmetric_division enabled="True">
    <extended_asymmetric_division_probability name1="type0" name2="type0" units="dimensionless">1.0</extended_asymmetric_division_probability>
  </extended_asymmetric_division>
</cycle>
```
**Note**: Order of `name1` and `name2` attributes does not matter. An ordered map on unordered pairs of integers is used to get/set the relevant probabilities.

# Behaviors
```sh
extended asymmetric division to <type1> and <type2>
```
**Note**: Order does not matter. An ordered map on unordered pairs of integers is used to get/set the relevant probabilities.

# Tolerance on the probability sum

The check that the probabilities sum to at most 1 allows a small excess, because probabilities meant to sum to exactly 1 land a hair over it in double precision — `0.11 + 0.33 + 0.56` is `1.000000000000000222`. That tolerance defaults to `1e-12` and is configurable, since how much slack a model needs depends on how many probabilities it sums and how they are produced. Probabilities driven by **rules** rather than XML literals have been seen to exceed 1 by around `8e-3`.

```xml
<options>
    <asymmetric_division_probability_tolerance>1e-6</asymmetric_division_probability_tolerance>
</options>
```

Absent, the default applies and nothing is printed. Present, the value is used and echoed at startup. A negative value is rejected at parse time rather than silently inverting the comparison. This only widens the band in which an over-specified distribution is accepted and absorbed — one that exceeds 1 beyond the tolerance still reports and exits, and nothing is rescaled.


drbergman:
Note, when merging into dev, this block will be needed to make the resume addition work in multicell ds:
```cpp
        if (create_cells)
		{
			pCell->phenotype.cycle.asymmetric_division.asymmetric_division_probabilities.clear();
			pCell->phenotype.cycle.asymmetric_division.asymmetric_division_probabilities.reserve(n_cell_types * (n_cell_types + 1) / 2);
		}
		for ( int i1 = 0; i1 < n_cell_types; i1++ )
		{
			for ( int i2 = i1; i2 < n_cell_types; i2++ )
			{
				fread(&dTemp, sizeof(double), 1, fp);
				if (debug_print)
				{ std::cout << " phenotype.cycle.asymmetric_division.set_asymmetric_division_probability(" << i1 << "," << i2 << ", " << dTemp << ")" << std::endl; }
				if (create_cells)
				{ pCell->phenotype.cycle.asymmetric_division.set_asymmetric_division_probability(i1, i2, dTemp); }
			}
		}
```

drbergman:
Update: this PR now supports a weights model for selecting asym div probabilities. the underlying data structures are used in both cases. A settings field now holds a boolean for which path to use. if using weights, the random number is not from `U(0,1)` but is on `U(0,total_weight)`. If total weight is 0, defaults to A -> A+A sym div the same way that the probability path does

PR #386, opened 2025-09-05
improve substrate csv initial conditions: omit voxels, fix out-of-bounds reads, revive headerless path

drbergman:
Currently, every voxel needs to be assigned a substrate concentration for every substrate column. Any substrates without a column are initialized as 0 everywhere, i.e., they ignore the XML.

Now, this PR allows for omitting voxels and having missing values in a row. Any missing voxel or missing value are set to 0. Thus,
```csv
x,y,z,sub1,sub2
20,0,0,1,
0,20,0,,1
```
1. initializes `sub1` to 0 everywhere except (20,0,0) where it is 1
2. initializes `sub2` to 0 everywhere except (0,20,0) where it is 1

Any other substrate would be 0 everywhere.

### DIFF
--- a/BioFVM/BioFVM_microenvironment.cpp
+++ b/BioFVM/BioFVM_microenvironment.cpp
@@ -1450,12 +1450,44 @@ void load_initial_conditions_from_matlab(std::string filename)
 	return;
 }
 
+// split a csv row into trimmed fields. A row of n commas always yields n+1 fields, matching how
+// substrate_csv_to_vector counts them, so a header and its data rows are measured the same way.
+static std::vector<std::string> split_substrate_csv_row(const std::string &line)
+{
+	std::vector<std::string> fields;
+	std::size_t start = 0;
+	while (true)
+	{
+		std::size_t comma = line.find(',', start);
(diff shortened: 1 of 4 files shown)


drbergman:
**Update:** folded in drbergman/PhysiCell#66. Beyond the omitted-voxels feature, this now fixes correctness and robustness problems that are present on `development` today.

Measured against `upstream/development` with libc++ bounds checking enabled. The value/column rows below use inputs that cover every voxel, so `development`'s own coverage check can't fire first and mask the cause:

| substrate IC csv input | `development` today | with this PR |
| --- | --- | --- |
| blank or whitespace-only row | out-of-bounds read | skipped |
| row with fewer columns than the header | out-of-bounds read | rejected, names the line |
| a field like `NA` | out-of-bounds read | rejected, names line and column |
| `inf` | out-of-bounds read | rejected, names line and column |
| `1.5abc` | **silently accepted as `1.5`** | rejected, names line and column |
| extra column on a row | **silently accepted, extra dropped** | rejected, names the line |
| position outside the domain | **silently written to an edge voxel** | rejected, reports position and domain |
| headerless file | rejected: `Wrong number of voxels supplied ... Found: 0` | loads correctly |

PR #388, opened 2025-09-09
remove checks for edge cases in boundary repulsion

drbergman:
- bottom line: at each mechanics time step, it will suffice to push the cell in a cardinal direction to increase the shortest distance to the boundary
- this is precisely what the current code currently does the vast majority of the time.
- now, it will do so for the previously defined edge cases
- in 2D sims, they looked for cells on line y=x
  - then used signs of x and y positions to determine which direction to push
  - no check for cells in upper left or lower right corners
  - the sign of the positions does not accurately convey which way to push (if all corners are negative, they would all push the same way)
- in 3D sims, they looked for cells on line x=y=z
  - similar issues but missing more cases (x=\pm y = \pm z)


drbergman:
@copilot can you rerun the checks, please??

MathCancer:
test comment

PR #392, opened 2025-10-15
CMake Support

asmelko:
Add CMakeLists.txt support to all sample projects and to PhysiBoSS intracellular project.

To try the new functionality, run:
```sh
# Configures project and creates build files in 'build' directory
# Optionally add -G Ninja for a more modern Ninja generator
cmake -B build -S .
# Build everything (all sample projects)
# Optionally specify --target TARGET to build only specific targets, such as biorobots
cmake --build build
```

The top-level CMakeLists.txt just adds various subdirectories containing other CMake files.
Directories `BioFVM`, `core` and `modules` define libraries, that will be (re)used to build sample projects.
In `sample_projects` directory, each sample project contains the same CMakeLists.txt which is responsible for building that project.
Intracellular sample projects are slightly different. They reference the library defined in `addons` directory. As of now, it is just boolean sample projects ported to CMake. See CMakeLists.txt in `addons/PhysiBoSS` for more info about how the external dependencies are fetched.

Currently, the biggest change compared to Make build is that CMake-built executables are located under the specified build directory. For example, biorobots executable and config files built under `build` will be located in `build/sample_projects/biorobots/` directory.


asmelko:
Added support for the remainder of addons. However, some of them are broken and can not compile even when using Make.

asmelko:
Also, a nice benefit is that I managed to compile PhysiCell on Windows without mingw. I just needed to change `create_directory` in `PhysiCell_settings.cpp`. But I needed to turn off MaBoSS compilation - the sources in there are not portable since they use UNIX routines heavily.

addons/libRoadrunner/src/librr_intracellular.h
```diff
@@ -16,8 +16,8 @@
 // #define STATIC_RRC
 // #include "rrc_api.h"
 // #include "rrc_types.h"
-#include "../roadrunner/include/rr/C/rrc_api.h"
-#include "../roadrunner/include/rr/C/rrc_types.h"
+#include "rrc_api.h"
+#include "rrc_types.h"
```
drbergman, line 20: I'm assuming that you've confirmed this still works? I don't know how C++ compilers resolve these paths.

.gitignore
```diff
@@ -7,10 +7,9 @@ addons/libRoadrunner/roadrunner/
 addons/libRoadrunner/roadrunner-win64-vs14-cp35m.zip
 addons/PhysiBoSS/MaBoSS/
 addons/PhysiBoSS/libMaBoSS-*.tar.gz
-biorobots
+build/
 cancer_immune_3D
 config/PhysiCell_settings-backup.xml
-heterogeneity
```
drbergman, line 13: Do we want to get rid of these lines from the .gitignore? `cancer_immune_3D` remained, so the inconsistency confuses the purpose of these changes. My vote would be to keep them in since it looks like we're not removing the Makefiles and thus not removing the current workflow for compiling and running projects, meaning these executables could end up in a user repo.

drbergman: and for any Windows users listening, do these even work if you have, e.g., `biorobots.exe`?

CMakeLists.txt
```diff
@@ -0,0 +1,68 @@
...
+# Set C++ standard and enable optimizations
+set(CMAKE_CXX_STANDARD 20)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -march=native")
+set(CMAKE_INTERPROCEDURAL_OPTIMIZATION TRUE)
+
+# Options to build addons
+option(PHYSICELL_BUILD_PHYSIBOSS "Build PhysiBoSS addon" ON)
+option(PHYSICELL_BUILD_LIBROADDRUNNER "Build libRoadrunner addon" OFF)
+option(PHYSICELL_BUILD_DFBA "Build dFBA addon" OFF)
```
drbergman, line 16: I don't know what these lines are doing, but it seems we're setting the default behavior to use PhysiBoSS (and not libRR or dFBA). If that's correct, I think we should default to all `OFF`.

drbergman (commented):
These changes look, on the whole, good to go. I cannot comment on the value they add since I don't understand how to make use of these. A few other comments below.

Looking forward to getting the power of HPC compute right at our fingertips!

.gitignore
asmelko: Ah, I see the inconsistency!

`biorobots` and `heterogeneity` in `.gitignore` were actually ignoring the whole directories in `sample_projects`. So when I added a CMakeLists.txt to those sample projects, these new files were (correctly) ignored by git. Now that the new CMake files are there, I can revert the changes in `.gitignore` (since already committed files can not be unignored), but I suggest we either remove `cancer_immune_3D` or we revert and prepend each sample project in `.gitignore` by `./` if we want to ignore just a binary in root.

addons/libRoadrunner/src/librr_intracellular.h
asmelko: Yes, it does.
In ODE Makefile, there is the correct include flag (https://github.com/asmelko/RePhysiCell/blob/cmake/sample_projects_intracellular/ode/ode_energy/Makefile#L40), so we can freely use relative paths.

CMakeLists.txt
asmelko: It is just that libRR and dFBA rely on files from `beta` directory, so I marked them turned off by default.

asmelko: But to be more specific about what these lines are doing - these are the options to conditionally add CMake projects into the whole build. Especially dFBA has various required dependencies, which may not be present on every system. Therefore, adding dFBA into CMake configuration step could fail the step for all other projects if the dependencies are not met. To solve this, it is a common CMake practice to provide options to conditionally allow/disallow building a subset of the project.

.gitignore
drbergman: Oh I see that! I tried the GitHub editor to open a PR on your branch with updates to the .gitignore, but GH threw a cryptic error. Turns out I could just directly commit to it 🤷 . That feels invasive. Sorry!

But I think this is what we want going forward. If down the line this runs into a merge conflict, blame me!

CMakeLists.txt
drbergman: OK, so the default build will include PhysiBoSS code even if the user doesn't use PhysiBoSS in their sims? I suppose at worst this just affects the size of the compiled binary then.

asmelko: it is a little more nuanced. CMake distinguishes 2 separate steps - configure step and build step. In configure step, all targets' dependencies are checked. Then, in build step, you choose which _configured_ target to build.

So to go back to our case, PhysiBoSS will be just _configured_ by default. In Make terms, PhysiBoss sample projects will be included into `make all` target. It is not that the executable size will be bigger, its just that cmake's variant of "make all" will build also all PhysiBoSS sample projects. Still, during cmake build step, you can choose to build only one target, such as biorobots, and nothing but biorobots will be build.

.gitignore
asmelko: I like the change!

CMakeLists.txt
drbergman: Oh cool! Sounds perfect for this use case, then.

rheiland:
I get:
```
$ cmake -B build -S .
-- The CXX compiler identification is AppleClang 17.0.0.17000013
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Could NOT find OpenMP_CXX (missing: OpenMP_CXX_FLAGS OpenMP_CXX_LIB_NAMES) 
-- Could NOT find OpenMP (missing: OpenMP_CXX_FOUND) 
CMake Warning at BioFVM/CMakeLists.txt:10 (message):
  OpenMP not found.  BioFVM will be built without OpenMP support.


CMake Error at addons/PhysiBoSS/CMakeLists.txt:31 (message):
  Unsupported OS: Darwin
```
On my Mac, CMake determines my default compiler is clang (which lacks OpenMP), whereas previously, our Makefiles would check for an env var PHYSICELL_CPP and use it if present (primarily intended for Mac brew g++).

asmelko:
> On my Mac, CMake determines my default compiler is clang (which lacks OpenMP), whereas previously, our Makefiles would check for an env var PHYSICELL_CPP and use it if present (primarily intended for Mac brew g++).

Yeah, so CMake supports this, but under a different env variable name.
You can either use `CC` and `CXX` env variables for setting C and C++ compilers respectively. So sth like this:
```
CXX=g++ cmake -B build -S .
```
Or a more modern but verbose variant:
```
cmake -B build -S . -DCMAKE_CXX_COMPILER=g++ 
```
Or you can brew install libomp and try if CMake can find it :)

rheiland:
Following up: @asmelko was able to fix some of the issues I was experiencing on a Mac build, so it all seems to be working for me - at least using a brew installed g++. I'm confident it could also be possible to build using Apple's native clang++ (with a brew installed libomp).

At least on a Mac, I think it should be possible to just use the  generated Makefile and resulting executable in the *root* dir (rather than down in /build) and, after modifying the `make reset` to *not* do `rm -f ./custom_modules/*`, be able to run the usual Makefile targets, e.g., `make reset`, `make template`, etc, from the root dir. More testing will be needed though.

Finally, on Windows, I encountered the same issues I had when I attempted to use cmake on a previous version of PhysiCell (but were all fixable then and likely now):
* Note that your own [free, community edition of the] Visual Studio C++ compiler may be different than mine:
```
C:\Users\randy\git\adam_cmake>cmake --version
cmake version 4.0.2
CMake suite maintained and supported by Kitware (kitware.com/cmake).
C:\Users\randy\git\adam_cmake> cmake -B build -S .
-- Building for: Visual Studio 17 2022
-- Selecting Windows SDK version 10.0.22000.0 to target Windows 10.0.26100.
-- The CXX compiler identification is MSVC 19.36.32532.0
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: C:/Program Files/Microsoft Visual Studio/2022/Community/VC/Tools/MSVC/14.36.32532/bin/Hostx64/x64/cl.exe - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Found OpenMP_CXX: -openmp (found version "2.0")
-- Found OpenMP: TRUE (found version "2.0")
-- Configuring done (10.0s)
-- Generating done (0.9s)
-- Build files have been written to: C:/Users/randy/git/adam_cmake/build
---------------------------------------

C:\Users\randy\git\adam_cmake>cmake --build build
MSBuild version 17.6.3+07e294721 for .NET Framework
  1>Checking Build System
  Building Custom Rule C:/Users/randy/git/adam_cmake/BioFVM/CMakeLists.txt
cl : command line  warning D9002: ignoring unknown option '-march=native' [C:\Users\randy\git\adam_cma
ke\build\BioFVM\BioFVM.vcxproj]
...
* numerous warnings, most of which are related to using `sprintf`
C:\Users\randy\git\adam_cmake\BioFVM\BioFVM_MultiCellDS.cpp(384,3): warning C4996: 'sprintf': 
This function or variable may be unsafe. Consider using sprintf_s instead.
...
conversion from 'size_t' to 'int',
...
* lots of these type of errors::
C:\Users\randy\git\adam_cmake\BioFVM\BioFVM_microenvironment.cpp(743,20): error C3016: 'i': 
index variable in OpenMP 'for' statement must have signed integral type

* a couple of these:
C:\Users\randy\git\adam_cmake\modules\PhysiCell_settings.cpp(363,17): error C3861: 'mkdir': 
identifier
 not found
```

asmelko:
> At least on a Mac, I think it should be possible to just use the generated Makefile and resulting executable in the root dir (rather than down in /build) and, after modifying the make reset to not do rm -f ./custom_modules/*, be able to run the usual Makefile targets, e.g., make reset, make template, etc, from the root dir. More testing will be needed though.

@rheiland There is another make/cmake functionality, which could be used for moving binaries to specified dirs (such as repo root), which is _install_. I can add install functionality for each target, which will "install" them into the repo root. I will be happy to hear your thoughts on this.

> Finally, on Windows, I encountered the same issues I had when I attempted to use cmake on a previous version of PhysiCell (but were all fixable then and likely now):

Let me start with that there are 2 compilers for windows: `cl` and relatively new `clang` (you can download both via Visual Studio Installer). I prefer the latter because it can read the "gnu" C++ flags, so you do not need to have complex os-dependent logic in Make for flags setting. Further, `cl` has only OpenMP 2.0 support, which is not enough for PhysiCell, while `clang` has the modern OpenMP 5.0.

Nevertheless, I added a commit that enables building PhysiCell for both Windows compilers. Firstly, I enable OpenMP only if a compiler supports versions >=4, and secondly, I solved `mkdir` usage in one source file. C++17 has support for filesystem, so I used that. But, since original Makefiles compile with C++11, I use filesystem only conditionally using ifdefs. That way, original makefiles are not touched.

Still, PhysiBoSS uses way too much UNIX routines, so that requires a separate PR for a direct Windows support. Therefore, you still need to configure cmake followingly, when using a Windows compiler:
```
cmake -B build -S . -DPHYSICELL_BUILD_PHYSIBOSS=OFF
```

rheiland:
> cl has only OpenMP 2.0 support, which is not enough for PhysiCell, while clang has the modern OpenMP 5.0

I don't recall having any issues with building an earlier PhysiCell release with cmake and `cl`, but could be mistaken. Just curious - what feature of OpenMP are we using that requires 5.0?

Regardless, the following tries to capture my current attempt to use cmake on Windows. It's a virgin machine - no C++ compiler, so I thought it might be possible to just install clang and then have cmake use it:
```
C:\Program Files\LLVM\bin>clang++.exe --version
clang version 21.1.4
Target: x86_64-pc-windows-msvc
Thread model: posix
InstalledDir: C:\Program Files\LLVM\bin
```
But cmake required installing ninja (and then editing my PATH env var to point to it):
```
C:\Users\heiland>ninja --version
1.13.1
```
Then, try to run cmake with clang:
```
C:\Users\heiland\dev\RePhysiCell-cmake\RePhysiCell-cmake>cmake -G Ninja -Bbuild -DCMAKE_C_COMPILER="C:/Program Files/LLVM/bin/clang.exe" -DCMAKE_CXX_COMPILER="C:/Program Files/LLVM/bin/clang++.exe"
-- The CXX compiler identification is Clang 21.1.4 with GNU-like command-line
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - failed
-- Check for working CXX compiler: C:/Program Files/LLVM/bin/clang++.exe
-- Check for working CXX compiler: C:/Program Files/LLVM/bin/clang++.exe - broken
CMake Error at C:/Program Files/CMake/share/cmake-4.1/Modules/CMakeTestCXXCompiler.cmake:73 (message):
  The C++ compiler

    "C:/Program Files/LLVM/bin/clang++.exe"

  is not able to compile a simple test program.

  It fails with the following output:

    Change Dir: 'C:/Users/heiland/dev/RePhysiCell-cmake/RePhysiCell-cmake/build/CMakeFiles/CMakeScratch/TryCompile-wmep8o'

    Run Build Command(s): C:/Users/heiland/bin/ninja.exe -v cmTC_c5e8e
    [1/2] C:\PROGRA~1\LLVM\bin\CLANG_~1.EXE   -O0 -g -Xclang -gcodeview -D_DEBUG -D_DLL -D_MT -Xclang --dependent-lib=msvcrtd -MD -MT CMakeFiles/cmTC_c5e8e.dir/testCXXCompiler.cxx.obj -MF CMakeFiles\cmTC_c5e8e.dir\testCXXCompiler.cxx.obj.d -o CMakeFiles/cmTC_c5e8e.dir/testCXXCompiler.cxx.obj -c C:/Users/heiland/dev/RePhysiCell-cmake/RePhysiCell-cmake/build/CMakeFiles/CMakeScratch/TryCompile-wmep8o/testCXXCompiler.cxx
    [2/2] C:\WINDOWS\system32\cmd.exe /C "cd . && C:\PROGRA~1\LLVM\bin\CLANG_~1.EXE -nostartfiles -nostdlib -O0 -g -Xclang -gcodeview -D_DEBUG -D_DLL -D_MT -Xclang --dependent-lib=msvcrtd -Xlinker /subsystem:console -fuse-ld=lld-link CMakeFiles/cmTC_c5e8e.dir/testCXXCompiler.cxx.obj -o cmTC_c5e8e.exe -Xlinker /MANIFEST:EMBED -Xlinker /implib:cmTC_c5e8e.lib -Xlinker /pdb:cmTC_c5e8e.pdb -Xlinker /version:0.0   -lkernel32 -luser32 -lgdi32 -lwinspool -lshell32 -lole32 -loleaut32 -luuid -lcomdlg32 -ladvapi32 -loldnames  && cd ."
    FAILED: [code=1] cmTC_c5e8e.exe
    C:\WINDOWS\system32\cmd.exe /C "cd . && C:\PROGRA~1\LLVM\bin\CLANG_~1.EXE -nostartfiles -nostdlib -O0 -g -Xclang -gcodeview -D_DEBUG -D_DLL -D_MT -Xclang --dependent-lib=msvcrtd -Xlinker /subsystem:console -fuse-ld=lld-link CMakeFiles/cmTC_c5e8e.dir/testCXXCompiler.cxx.obj -o cmTC_c5e8e.exe -Xlinker /MANIFEST:EMBED -Xlinker /implib:cmTC_c5e8e.lib -Xlinker /pdb:cmTC_c5e8e.pdb -Xlinker /version:0.0   -lkernel32 -luser32 -lgdi32 -lwinspool -lshell32 -lole32 -loleaut32 -luuid -lcomdlg32 -ladvapi32 -loldnames  && cd ."
    lld-link: error: could not open 'kernel32.lib': no such file or directory
    lld-link: error: could not open 'user32.lib': no such file or directory
    ... (lots more missing libs)
```

Unsure what to make of this, so I installed Visual Studio Community edition (~5GB). Then re-try the cmake with clang build:
```
C:\Users\heiland\dev\RePhysiCell-cmake\RePhysiCell-cmake>rm -rf build

C:\Users\heiland\dev\RePhysiCell-cmake\RePhysiCell-cmake> cmake -G Ninja -B build -DCMAKE_C_COMPILER="C:/Program Files/LLVM/bin/clang.exe" -DCMAKE_CXX_COMPILER="C:/Program Files/LLVM/bin/clang++.exe"
-- The CXX compiler identification is Clang 21.1.4 with GNU-like command-line
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: C:/Program Files/LLVM/bin/clang++.exe - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Could NOT find OpenMP_CXX (missing: OpenMP_CXX_FLAGS OpenMP_CXX_LIB_NAMES) (Required is at least version "4")
-- Could NOT find OpenMP (missing: OpenMP_CXX_FOUND) (Required is at least version "4")
CMake Warning at BioFVM/CMakeLists.txt:10 (message):
  OpenMP not found.  BioFVM will be built without OpenMP support.


-- Configuring done (11.3s)
-- Generating done (0.8s)
CMake Warning:
  Manually-specified variables were not used by the project:

    CMAKE_C_COMPILER

-- Build files have been written to: C:/Users/heiland/dev/RePhysiCell-cmake/RePhysiCell-cmake/build
```

And then:
```
C:\Users\heiland\dev\RePhysiCell-cmake\RePhysiCell-cmake> cmake --build build
… (skip over warnings)
4 warnings generated.

[36/128] Building CXX object modules/CMakeFiles/PhysiCell-modules.dir/PhysiCell_settings.cpp.obj
FAILED: [code=1] modules/CMakeFiles/PhysiCell-modules.dir/PhysiCell_settings.cpp.obj
C:\PROGRA~1\LLVM\bin\CLANG_~1.EXE   -march=native -O0 -g -Xclang -gcodeview -D_DEBUG -D_DLL -D_MT -Xclang --dependent-lib=msvcrtd -std=gnu++20 -flto=thin -MD -MT modules/CMakeFiles/PhysiCell-modules.dir/PhysiCell_settings.cpp.obj -MF modules\CMakeFiles\PhysiCell-modules.dir\PhysiCell_settings.cpp.obj.d -o modules/CMakeFiles/PhysiCell-modules.dir/PhysiCell_settings.cpp.obj -c C:/Users/heiland/dev/RePhysiCell-cmake/RePhysiCell-cmake/modules/PhysiCell_settings.cpp
C:/Users/heiland/dev/RePhysiCell-cmake/RePhysiCell-cmake/modules/PhysiCell_settings.cpp:368:6: error: no member named 'create_directories' in namespace 'std::filesystem'; did you mean simply 'create_directories'?
  368 |         if (std::filesystem::create_directories(path, ec)) {
      |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |             create_directories
C:/Users/heiland/dev/RePhysiCell-cmake/RePhysiCell-cmake/modules/PhysiCell_settings.cpp:339:6: note: 'create_directories' declared here
  339 | bool create_directories(const std::string &path)
      |      ^
C:/Users/heiland/dev/RePhysiCell-cmake/RePhysiCell-cmake/modules/PhysiCell_settings.cpp:368:48: error: too many arguments to function call, expected single argument 'path', have 2 arguments
  368 |         if (std::filesystem::create_directories(path, ec)) {
      |             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~       ^~
C:/Users/heiland/dev/RePhysiCell-cmake/RePhysiCell-cmake/modules/PhysiCell_settings.cpp:339:6: note: 'create_directories' declared here
  339 | bool create_directories(const std::string &path)
      |      ^                  ~~~~~~~~~~~~~~~~~~~~~~~
2 errors generated.
[41/128] Building CXX object modules/CMakeFiles/P...ell-modules-no-addon.dir/PhysiCell_pugixml.cpp.ob
ninja: build stopped: subcommand failed.
```

asmelko:
@rheiland Thanks a lot for being a testing dummy! :)

> I don't recall having any issues with building an earlier PhysiCell release with cmake and cl, but could be mistaken. Just curious - what feature of OpenMP are we using that requires 5.0?

I am adding OpenMP >=4.0 requirement only for this kind of errors that we get when compiling with `cl`:
```
C:\Users\randy\git\adam_cmake\BioFVM\BioFVM_microenvironment.cpp(743,20): error C3016: 'i': 
index variable in OpenMP 'for' statement must have signed integral type
```
As my optimizations porting will be advancing, I may experiment with lowering the OpenMP version requirement and fixing some of these errors if it brings us interesting Windows portability. 

Finally, I added a missing `version` include, now Windows build should compile. I added GitHub action to make sure of that.

rheiland:
Using your latest code, I now get this and will try to find reasons/alternatives, but welcome comments:
```
C:\Users\heiland\dev\RePhysiCell-cmake>cmake -G Ninja -B build -DCMAKE_C_COMPILER="C:/Program Files/LLVM/bin/clang.exe" -DCMAKE_CXX_COMPILER="C:/Program Files/LLVM/bin/clang++.exe"
-- The CXX compiler identification is Clang 21.1.4 with GNU-like command-line
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: C:/Program Files/LLVM/bin/clang++.exe - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Could NOT find OpenMP_CXX (missing: OpenMP_CXX_FLAGS OpenMP_CXX_LIB_NAMES) (Required is at least version "4")
-- Could NOT find OpenMP (missing: OpenMP_CXX_FOUND) (Required is at least version "4")
CMake Warning at BioFVM/CMakeLists.txt:10 (message):
  OpenMP not found.  BioFVM will be built without OpenMP support.


-- Configuring done (11.0s)
-- Generating done (0.6s)
CMake Warning:
  Manually-specified variables were not used by the project:

    CMAKE_C_COMPILER


-- Build files have been written to: C:/Users/heiland/dev/RePhysiCell-cmake/build
```

asmelko:
@rheiland If you are referring to CMake Warning for unused variables, I would not bother by that too much. But if you mean a not found OpenMP support, that is concerning. I installed clang and LLVM toolset via visual studio installer and for me, OpenMP is found:
```
[proc] Executing command: "C:\Program Files\CMake\bin\cmake.EXE" -DCMAKE_EXPORT_COMPILE_COMMANDS:BOOL=TRUE "-DCMAKE_C_COMPILER:FILEPATH=C:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools\VC\Tools\Llvm\bin\clang.exe" "-DCMAKE_CXX_COMPILER:FILEPATH=C:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools\VC\Tools\Llvm\bin\clang.exe" --no-warn-unused-cli -S C:/Users/adams/Desktop/RePhysiCell -B c:/Users/adams/Desktop/RePhysiCell/build -G Ninja
[cmake] Not searching for unused variables given on the command line.
[cmake] -- The CXX compiler identification is Clang 19.1.5 with GNU-like command-line
[cmake] -- Detecting CXX compiler ABI info
[cmake] -- Detecting CXX compiler ABI info - done
[cmake] -- Check for working CXX compiler: C:/Program Files (x86)/Microsoft Visual Studio/2022/BuildTools/VC/Tools/Llvm/bin/clang.exe - skipped
[cmake] -- Detecting CXX compile features
[cmake] -- Detecting CXX compile features - done
[cmake] -- Found OpenMP_CXX: -fopenmp=libomp (found suitable version "5.1", minimum required is "4")
[cmake] -- Found OpenMP: TRUE (found suitable version "5.1", minimum required is "4")
[cmake] -- Configuring done (4.3s)
[cmake] -- Generating done (0.4s)
[cmake] -- Build files have been written to: C:/Users/adams/Desktop/RePhysiCell/build
```

Maybe you also have a more behaving clang compiler version under `BuildTools` directory too?

asmelko:
Added GitHub actions for Building and Testing using CMake. Everything passes on Windows, MacOS, Ubuntu.

heberlr:
I tested this on my macOS and it worked. However, I think we should highlight the use of `CXX=$PHYSICELL_CPP cmake -B build -S .` if `$PHYSICELL_CPP` is defined, similar to how it is done in the MakeFile. This should help users who follow the installation tutorials.

asmelko:
Imagine a common physicell flow:
```sh
make biorobots # configures makefiles to build biorobots
make # builds biorobots
```
After these 2 steps, one has biorobots exe ready in the root of the project with config files, custom module and main copied along.

Now this is the proposed flow:
```sh
# runs cmake configure step, ordering that "build" is the build directory and that the top-level cmake file is in "." this directory
cmake -B build -S . 
# runs cmake build step, looking for build instructions in "build" dir, building "only" biorobots
cmake --build build --target biorobots
```

Although it looks same, there are already major differences:
1. the fist cmake step configures **all** the projects. There is little to no cost to it. It is robust enough to hold multiple "make targets" (cmake targets) in a single build tree. You can do `cmake --build build --target help` to see all available targets.
2. the second cmake step builds **only** biorobots target. If you omit `--target biorobots` it will build **all** targets, which indeed takes time. Still, BioFVM, core, and modules are built just once and reused across multiple targets.
3. The consequence of this is that the built executables are stored in the same tree structure as the source files, under the specified build directory. So the biorobots exe will be in `build/sample_projects/biorobots/biorobots` instead of `./biorobots`. This is a more organized approach, especially when dealing with multiple targets. The custom modules and main sources are not copied around; they are built in place within the build directory structure. Config directory is still copied to the build tree for each target, so that running `./biorobots` from its build dir still works as expected.

**Question 1:** Is this new executable location acceptable? There is a notion of "cmake install" that can copy the built executables to a specified location, but that is an additional step and commonly used to install "everything". Not sure if we can install just target. If needed, I can explore this.

Next, to change the compiler, one changes to (first) configure step to:
```sh
cmake -B build -S . -DCMAKE_CXX_COMPILER=g++-11
```
This will use g++-11 for all targets built in this build tree.

**Question 2:** Is this new flow acceptable? Setting the compiler like this is standard in the CMake world. It offers more flexibility, especially when dealing with multiple targets and configurations. However, I think it is possible to fetch the content of PHYSICELL_CPP env var and use it to set CMAKE_CXX_COMPILER within the cmake files, if that is preferred.

Next, I copied the github test and build workflows and adapted them to this new flow. They are working fine.

**Question 3:** Is this new workflow acceptable for the CI as well? It is more robust and scalable, but I want to ensure it fits within your existing processes.

Next, I added conditional inclusion of dfba and libroadrunner in the configure step. This means that the addons *wont even configure (i.e., show as buildable targets)* unless the user explicitly requests them by setting `-DPHYSICELL_BUILD_LIBROADDRUNNER=ON` and/or `-DPHYSICELL_BUILD_DFBA=ON` during the configure step. This keeps the default configuration lean and avoids unnecessary fails that are caused by missing dependencies that some of these addons required. So if a user does not want to use the addon, they dont have to install its dependencies.

**Question 4:** Is this conditional inclusion acceptable? Is the naming acceptable? 

Finally, let me describe the CMake architecture briefly:
- Top-level CMakeLists.txt:
    - Lines 8 to 11 set up project-wide C++ flags, they include C++20 compliance, -march=native and link-time optimization (LTO).
    - Lines 13 to 16 set up options to include/exclude certain addons (libroadrunner and dfba).
    - Lines 18 to 22 define main PhysiCell libraries: BioFVM, core, modules.
    - Remainder include sample projects one by one.
- Each sample project has its own CMakeLists.txt that defines its executable target, its source files (custom modules and main), and links it against the necessary libraries (core, BioFVM, and any other modules).
- Its worthwile looking at addons CMakeLists.txt files. They subsidize the python scripts that fetch the dependencies, such as maboss shared library.

All sample-projects-related CMakeLists.txt are almost identical. The ideal flow of adding a new sample project would be just to copy `sample_projects/CmakeLists-default.txt` to `sample_projects/<new_project>/CMakeLists.txt`, and then adjust the top-level CMakeLists.txt to include the new project.

**Question 5:** Is this CMake architecture acceptable? I think it could be possible to have just one CMakeLists.txt for all sample projects that loops over all subdirectories, but that would be more complex and less explicit. I opted for explicitness here.

**Question 6:** Are C++ flags set in the top-level CMakeLists.txt acceptable?

**Question 7:** Is minimal CMake version 3.26 acceptable? Maybe for some systems it is too new? I can try to lower it if needed.

Please all, try this new flow and provide feedback on the above questions. If all is acceptable, let's merge.

vincent-noel:
Q1: I would try to do a cmake install, to put the executable in the root folder. At least during a transition period, not to brake PhysiCell Studio default behavior.

Q2: Ok for me. This is the default usage of cmake, let's keep it standard.

Q3: We will have to adapt the CI, create a new one for cmake. That way, we test both, and make sure both work. EDIT: you already did create new ones, so perfect ! 

Q4: Yes. Maybe PhysiBoSS should also be off by default ? If I remember correctly, this is still not working for Windows MSVC.

Q5: Yes, this is the good structure for cmake. For the addons, I'll work on the PhysiBoSS one to update MaBoSS version, and ensure MSVC compatibility. 

Q6: Do we want to have more verbosity, such as Wall or pedantic ? Or is it by default ?

Q7: Oldest Ubuntu LTS which is still maintained (22.04) ships 3.22.1. So maybe make it compatible with that one ?

rheiland:
Suppose a modeler is C++-savvy and includes additional custom C++ files besides custom.cpp. We will want clear instructions for how they should modify the cmake file(s).

rheiland:
Testing on Windows:

```
C:\Users\randy\git\adam_physicore_integration> cmake -DPHYSICELL_USE_PHYSICORE_BIOFVM=ON -DCMAKE_BUILD_TYPE=Release  -B build
-- Building for: Visual Studio 17 2022
-- Selecting Windows SDK version 10.0.22000.0 to target Windows 10.0.26100.
-- The CXX compiler identification is MSVC 19.36.32532.0
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: C:/Program Files/Microsoft Visual Studio/2022/Community/VC/Tools/MSVC/14.36.32532/bin/Hostx64/x64/cl.exe - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Setting BioFVM_DIR to C:/Users/randy/git/adam_physicore_integration/build/_deps/physicore-src/lib/cmake/BioFVM
-- Could NOT find OpenMP_CXX: Found unsuitable version "2.0", but required is at least "4" (found -openmp)
-- Could NOT find OpenMP (missing: OpenMP_CXX_FOUND) (Required is at least version "4")
CMake Warning at external/physicore-adapter/CMakeLists.txt:37 (message):
  OpenMP not found.  BioFVM will be built without OpenMP support.
...
```
* Is there a requirement for a more recent version of VS/MSVC? 
* Do we know what PhysiCell is doing with OpenMP to require version 4?

I'm just trying to minimize what a modeler needs to do to build.

asmelko:
@rheiland, I investigated why MSVC OpenMP 2.0 could not compile PhysiCell and it was a simple fix. I just had to use signed integers in omp pragmas.

asmelko:
Regarding the documentation - point me to the most suitable place in the repo, and I can draft something!

PR #401, opened 2025-12-15
BioFVM Interface

asmelko:
The PR introduces 2 main interface classes:
- `Microenvironment_Interface`
- `Basic_Agent_Interface`

These interfaces are used throughout the rest of PhysiCell code instead of the _concrete BioFVM classes_ so that we can support multiple BioFVM implementations. 

As of now, the interfaces are very close to 1-to-1 copies of the concrete classes. In the next PR, which introduces a new BioFVM implementation, we will refine the interfaces further by removing internal methods.

`BioFVM_Implementation` class is the _pseudo-factory_ for the new implementation: The new implementation will have to implement the methods of this class that will operate with the concrete classes of the 2 aforementioned interfaces.

The code change is big, but it is mainly just a necessary syntactic change to support proper interface isolation. However, here are the major logic changes:
-  molecular and secretion phenotypes now **dont own the data** (vector of rates). The code has been refactored such that the data is owned by BioFVM, which removes the confusing synchronization code and leaves the implementation details to the BioFVM implementater for the best possible performance gain. 
- some BioFVM-related functions that are reading settings.xml have been moved from `modules` to `BioFVM` directory
- all sample-projects' main files now start with `BioFVM_Implementation::initialize_microenvironment();` that will initialize a BioFVM implementation for the further modification and use
- `all_basic_agents` are now only accessible via `BioFVM_Implementation::get_all_basic_agents` for the sake of hiding implementation details


BioFVM/BioFVM_basic_agent.cpp
```diff
@@ -301,7 +311,7 @@ void Basic_Agent::set_total_volume(double volume)
 	volume_is_changed = true;
 }
 
-double Basic_Agent::get_total_volume()
+double& Basic_Agent::get_total_volume()
```
drbergman, line 314: There is a specialized setter for volume that sets the `volume_is_changed` flag. Are we sure we need to access this by reference?

drbergman:
```cpp
void Basic_Agent::set_total_volume(double volume)
{
	this->volume = volume;
	volume_is_changed = true;
}
```

core/PhysiCell_cell_container.cpp
```diff
@@ -122,17 +121,6 @@ void Cell_Container::update_all_cells(double t)
...
-
-	#pragma omp parallel for 
-	for( int i=0; i < (*all_cells).size(); i++ )
-	{
-		if( (*all_cells)[i]->is_out_of_domain == false )
-		{
-			(*all_cells)[i]->phenotype.secretion.advance( (*all_cells)[i], (*all_cells)[i]->phenotype , diffusion_dt_ );
-		}
-	}
-
```
drbergman, line 135: Where does this code get moved to? I can't find a call to `secretion.advance` or an instance of `advance(`

asmelko: that one is now in `get_microenvironment_i()->simulate_time_step( diffusion_dt );` added in `main.cpp` of sample projects. see `Microenvironment_Adapter::simulate_time_step` for the implementation. I added this new function to groups BioFVM simulation parts instead of having them dispersed in `main.cpp` via `microenvironment.simulate_diffusion_decay` and in `Physicell_cell_container.cpp` via `((Cell_Container *)microenvironment.agent_container)->update_all_cells`

this one is definitely something I will need to refactor so that we do not modify sample projects for default biofvm impl

BioFVM/BioFVM_basic_agent.cpp
asmelko: ah so this is necessary because now that `Basic_Agent::get_total_volume()` is virtual, it is clashing with `Cell::get_total_volume()`, which returns the reference. 

So we either unify `Cell`'s and `Basic_Agents`'s `get_total_volume` to return `double&` or we can actually remove `Cell::get_total_volume` (since its deprecated) and keep `double`.

I am open to ideas

asmelko:
As requested, I make the BioFVM implementation choice dynamic. 
The entry point to the implementations is `BioFVM/BioFVM_implementation.h` - its a factory pattern that returns implementation-related objects. Currently, there is just the original implementation residing in `BioFVM/BioFVM_legacy_implementation.h`. 

The dynamical choice is currently not very dynamic, it lives in `load_PhysiCell_config_file` function and now it is hard-coded to always use the original BioFVM.

Further, I reverted some parts of the code so that sample_projects changes are minimized. Still, there are some changes mainly related to the the basic agent abstraction. A consequence of this is that the sample projects are now skipping the BioFVM_implementation indirection and are hard-coded to use the original biofvm.

rheiland:
I did a preliminary test comparing the `cancer-immune-sample` (3D tumor) project and seem to see this branch taking about 1.4 times longer to run than the 1.14.2 release. Using brew g++-15 on Apple M1 and using 4 threads. I've not done any profiling or compared results yet, but the cell counts are reasonably close.

rheiland:
After discussing this further with Adam, I pulled the latest from this repo (after he rebased on PhysiCell development branch), built using cmake, and compared with the PhysiCell development branch - using the same `cancer-immune-sample` model, same Mac M1 machine, same 4 threads and still seeing about 1.4x difference in wall clock sim time (below). Hopefully others will be able to report their results.
```
<this-repo>/build/sample_projects/cancer_immune$ tail tumor3D.out
...
current simulated time: 10800 min (max: 30240 min)
total agents: 84083
interval wall time: 0 days, 1 hours, 12 minutes, and 24.0433 seconds
total wall time: 0 days, 10 hours, 53 minutes, and 57.2977 seconds

vs.
~/dev/PhysiCell_dev$ tail tumor3D.out
...
current simulated time: 10800 min (max: 30240 min)
total agents: 83264
interval wall time: 0 days, 0 hours, 46 minutes, and 51.491 seconds
total wall time: 0 days, 7 hours, 36 minutes, and 56.3602 seconds
```

PR #406, opened 2026-02-09
Update libroadrunner setup

rheiland:
* cleanup and update the `beta/setup_libroadrunner.py` script
   * we now get more recent headers and libs (.zip bundles) from `https://github.com/PhysiCell-Tools/intracellular_libs/tree/main/ode` instead of very old ones from sourceforge
   * these are minimal versions of those provided at `https://github.com/sys-bio/roadrunner/releases` (but probably not the very latest release)
 * cleanup and fix the Makefile for the `ode-energy-sample` project to use the new directory structure in these new .zip bundles
 * tested on Windows, Mac (both Silicon and Intel), and Linux (Ubuntu 24.04)

### DIFF
--- a/addons/libRoadrunner/CMakeLists.txt
+++ b/addons/libRoadrunner/CMakeLists.txt
@@ -36,7 +36,7 @@ FetchContent_Declare(
 FetchContent_MakeAvailable(roadrunner)
 
 target_include_directories(libRoadrunner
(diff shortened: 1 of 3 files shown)


drbergman:
@vincent-noel and @rheiland will discuss how to make this automated and stripped out the unnecessary bits of librr (we only need the C API)

PR #409, opened 2026-02-16
Fix(cell-neighbors): update max_cell_interactive_distance_in_voxel on position change

heberlr:
Based on  the discussion in PR #397, we found that `max_cell_interactive_distance_in_voxel` was updating too infrequently. This resulted in stale data and misleading neighbor list calculations.

This PR ensures that max_cell_interactive_distance_in_voxel is updated every time a cell's position is updated. This guarantees that neighbor tracking based on voxel neighborhoods remains accurate and reflects the most recent state of the simulation.


drbergman:
Different than what we initially discussed, but also effective! It looks like we already do a check for the mechanics voxel index changing in `void Cell::update_voxel_in_container()`. We could put this code there so that we only perform these checks and updates only if the cell has changed mechanics voxel.

vincent-noel:
Thanks for looking into this !
I'm a bit out of the loop, is it supposed to completely fix the neighbors issue in #397 ? It looks like I still have some inconsistencies in neighbors list with this fix.

heberlr:
> Thanks for looking into this ! I'm a bit out of the loop, is it supposed to completely fix the neighbors issue in #397 ? It looks like I still have some inconsistencies in neighbors list with this fix.

Thanks @vincent-noel. This a fix to update `max_cell_interactive_distance_in_voxel` after a cell changes its mechanical voxel, as this was generating inconsistencies in the neighbors list. 

In my simple test, this fix resolved the inconsistency, but I am open to testing it in different models. Could you share the model or provide an isolated case where occurs this inconsistency occurs? I'd be happy to take a look.

drbergman:
> Thanks for looking into this ! I'm a bit out of the loop, is it supposed to completely fix the neighbors issue in #397 ? It looks like I still have some inconsistencies in neighbors list with this fix.

Does inconsistency here still just mean asymmetric neighbor relationships? If so, have you checked if one of the cells is outside the domain? #410 may help. Curious if this + that could resolve?

drbergman:
just reread #410 and realized it stuck in my head as about out of domain. in fact, it is for unmovable cells.

Nonetheless, the question is if one of the cells you're finding in these asymmetric pairs is outside the domain. We don't maintain their neighbor list so asymmetry is expected there

drbergman:
@vincent-noel will make a new PR that will be accepted. see future comment for which PR to merge. tag @MathCancer on that PR

PR #410, opened 2026-02-17
Include neighbor lists and pressure computation in non-movable cells.

heberlr:
Prior to this PR, non-movable cells did not have their own neighbor lists updated, even though they could still appear in the neighbor lists of movable cells. As a result, simple pressure was not computed for non-movable cells, and their mechanical state was incomplete.

This PR explicitly sets the velocity of non-movable cells to zero, ensuring they remain mechanically immobile while still participating fully in neighbor detection and pressure calculations. Non-movable cells now maintain updated neighbor lists and properly contribute to mechanical interactions without exhibiting motion.


drbergman:
@vincent-noel will make a fresh PR for this. same as in #409

PR #411, opened 2026-02-18
update the if-query so that it spans all possible contact with cells cases.

elmbeech:
the possible cases: 
```
n:  contact with cell type
n + 1:  contact with live cell
n + 2:  contact with dead cell  <= here the current implementation stops
n + 3:  contact with apoptotic cell  <= bug fix
n + 4:  contact with necrotic cell  <= bug fix
n + 5:  contact with other dead cell  <= bug fix
```

### DIFF
--- a/core/PhysiCell_signal_behavior.cpp
+++ b/core/PhysiCell_signal_behavior.cpp
@@ -79,6 +79,8 @@ std::map<std::string,int> behavior_to_int;
 std::map<int,std::string> int_to_signal; 
 std::map<int,std::string> int_to_behavior; 
 
+int contact_with_built_in = 5;  // live, dead, apoptotic, necrotic, other
+
 void setup_signal_behavior_dictionaries( void )
 {
 	extern std::unordered_map<std::string,int> cell_definition_indices_by_name; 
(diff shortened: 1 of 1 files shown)


drbergman (changes requested):
great catch! Seems like the right fix. I found two other bugs to fix:

1. At the end of `get_cell_contact_signals`
```cpp
std::string search_for = "contact with " + cell_definitions_by_type[0]->name; 
static int scaling_start_index = find_signal_index( search_for ); 
for( int i=0; i < n+2 ; i++ )
{ output[i] /= signal_scales[scaling_start_index+i]; }
```
we should loop until `i < n+5`

2. In this block in `get_selected_signals`
```cpp
for( int i=0; i < indices.size() ; i++ )
{
	int ind = indices[i]; 
	if( ind >= contact_start_index && ind < contact_start_index+n+2)
	{ signals[i] = contact_signals[ind-contact_start_index]; }
	else
	{ signals[i] = get_single_signal( pCell , ind ); }
}
```
the check should instead be `ind < contact_start_index+n+5` (the `5` is the difference). @vincent-noel It looks like PhysiBoSS uses `get_selected_signals`. Ultimately, that falls back to `get_single_signal` which would spam a warning message if this bug was affecting your users. So no extra concern here for this function.

elmbeech:
@drbergman  , i wonder, would it maybe not be better to make constant:
```
int contact_with = 5;
```
and then just use this constant everywhere.
so the value has to be set only once.
(i am not sure about what would be the best name for this constant.)

drbergman:
sure! Makes sense as all these changes are within one file.

elmbeech:
@drbergman  finally, i came back to this.
i added a contact_with_built_in variable, as discussed.

drbergman:
@MathCancer agreed to merge

PR #415, opened 2026-03-31
Combining intracellular models with a single model

vincent-noel:
Hi all !

This is a PR on something I've been thinking about for some time: the possibility of having multiple types of intracellular models, and to have them interact. To achieve this, we need to modify a bit the implementation of Intracellular, mainly to change the variable in Phenotype, from an Intracellular pointer to a vector of Intracellular pointers.

This leads to a few more changes, such as having the pre/post intracellular update function belonging to the Intracellular object, and to add the Intracellular to a few method parameters (to be sure to have the "good" one). 

Here I'm providing a quick example I build in the last days, which has an SBML ODE model for the TNF receptor binding and endocytosis and a Boolean model for the cell fate. To build it, use `make spheroid_tnf`. 

For the ODE model, I'm using what I think is the latest implementation of roadrunner, based on #362 and #406. I had to modify a few things, such as introducing the Intracellular* parameter in many places, and I also changed the way the species values are modified before the update. So before thinking about merging this, we should focus on #362 and #406, and I can make a separate PR to have my changes too. 

I am also working on example using the latest version of dFBA which is being developed at BSC by @marcorusc and @migp11, but I'll wait for them to submit their PR.

Let me know what you think !


drbergman:
I think this is really cool! We should have a conversation around next updates to `Intracellular` together since, as you pointed out, this interacts in non-trivial ways with other open PRs. This feels like a use-case for a GitHub project...

migp11:
This sounds great! Together with @marcorusc  and @othmya, we are finishing the physicell-dfba paper, which includes several different use cases that can be 
@vincent-noel have you reimplemented the TNF receptor model in SBML from this publication?
- [https://www.frontiersin.org/journals/molecular-biosciences/articles/10.3389/fmolb.2022.836794/full](https://www.frontiersin.org/journals/molecular-biosciences/articles/10.3389/fmolb.2022.836794/full)
That is a very nice test case!
Regarding the implementation of a hybrid model, including a Boolean model and an FBA-based metabolic model, I've been thinking and got some ideas based on @othmya last paper
- [https://www.nature.com/articles/s41540-026-00669-4
](https://www.nature.com/articles/s41540-026-00669-4
)and another paper we did on AGS metabolism
- [https://www.nature.com/articles/s41540-025-00586-y](https://www.nature.com/articles/s41540-025-00586-y)
The nice thing about this potential use case is that there is already a Boolean model and a metabolic one.
Open to discuss the topic :)
Best

vincent-noel:
@migp11 Indeed, this is your TNF receptor model. That was the simplest example I could think of, and it worked. 

About the AGS model, that's a nice idea. I have some reading to do, and then let's talk about this.

addons/libRoadrunner/src/librr_intracellular.cpp
```diff
@@ -15,13 +15,13 @@ void RoadRunnerMapping::initialize_mapping( void )
         index = behavior_ind;
         if (use_for_input)
         {
-            value_map = [this](PhysiCell::Cell *pCell)
-            { pCell->phenotype.intracellular->set_parameter_value(this->sbml_species, PhysiCell::get_single_behavior(pCell, this->index)); };
+            value_map = [this](PhysiCell::Cell *pCell, RoadRunnerIntracellular *intracellular)
+            { intracellular->set_parameter_value(this->sbml_species, PhysiCell::get_single_behavior(pCell, this->index)); };
```
drbergman: I'll be curious to understand why this anonymous function needs this second argument.

core/PhysiCell_cell.cpp
```diff
@@ -666,10 +666,12 @@ Cell* Cell::divide( )
 	// pAttackTarget was cleared above, so the daughter inherits a NULL one
 	child->phenotype = phenotype; 
 
-    if (child->phenotype.intracellular){
-        child->phenotype.intracellular->start();
-		child->phenotype.intracellular->inherit(this);
+	for (size_t i=0; i < child->phenotype.intracellulars.size(); i++)
+	{
+		child->phenotype.intracellulars[i]->start();
```
drbergman, line 671: just noting here that if we need to decide how to proceed with #425. the call to `start()` here would not be necessary. and I think we could make the `inherit` call unnecessary.

core/PhysiCell_cell.cpp
```diff
@@ -1137,8 +1139,8 @@ Cell* create_cell( Cell_Definition& cd )
 	pNew->functions = cd.functions; 
 	
 	pNew->phenotype = cd.phenotype; 
-	if (pNew->phenotype.intracellular)
-		pNew->phenotype.intracellular->start();
+	for (auto* intracellular: pNew->phenotype.intracellulars)
+		intracellular->start();
```
drbergman, line 1143: We're confident we don't need to `inherit` here? I honestly don't know.

core/PhysiCell_cell.cpp
```diff
@@ -3192,13 +3194,13 @@ Cell_Definition* initialize_cell_definition_from_pugixml( pugi::xml_node cd_node
...
-			if (pParent != NULL && pParent->phenotype.intracellular != NULL) {
-				pCD->phenotype.intracellular->initialize_intracellular_from_pugixml(node);
+			if (pParent != NULL && pParent->phenotype.intracellulars.size() > 0) {
+				pCD->phenotype.intracellulars[0]->initialize_intracellular_from_pugixml(node);
 				
 			// Otherwise we need to create a new one
 			} else {
 				MaBoSSIntracellular* pIntra = new MaBoSSIntracellular(node);
-				pCD->phenotype.intracellular = pIntra->getIntracellularModel();
+				pCD->phenotype.intracellulars.push_back(pIntra->getIntracellularModel());
```
drbergman, line 3203: This code diff is hard for me to understand.
```
If the parent has an intracellular model,
    initialize the first intracellular of this cell def from the node.
else
    push a new one onto the end
```
Assuming that `pCD` inherits its intracellulars from `pParent`, we can't guarantee they share the same intracellular type. Also, why are we only updating the first entry here? I'll look for where we update further entries below.

core/PhysiCell_cell.cpp
```diff
@@ -3225,22 +3233,98 @@ Cell_Definition* initialize_cell_definition_from_pugixml( pugi::xml_node cd_node
...
+			// If it has already be copied
+			if (pParent != NULL && pParent->phenotype.intracellulars.size() > intracellular_count) {
+				pCD->phenotype.intracellulars[intracellular_count]->initialize_intracellular_from_pugixml(intracellular);
+			// Otherwise we need to create a new one
+			} else {
+				PhysiCelldFBA::dFBAIntracellular* pIntra = new PhysiCelldFBA::dFBAIntracellular(intracellular);
+				pCD->phenotype.intracellulars.push_back(pIntra->getIntracellularModel());
+			}
+		}
+#endif
```
drbergman, line 3316: ah! This is what I was looking for. so what's the part above meant for then?

core/PhysiCell_phenotype.cpp
```diff
@@ -1214,18 +1217,45 @@ Phenotype& Phenotype::operator=(const Phenotype &p ) {
...
+	this->pre_update_intracellular = intracellular->pre_update_intracellular;
+	this->post_update_intracellular = intracellular->post_update_intracellular;
+}
+	
+Intracellular* Intracellular::operator=(const Intracellular *intracellular ) { 
+
+	this->pre_update_intracellular = intracellular->pre_update_intracellular;
+	this->post_update_intracellular = intracellular->post_update_intracellular;
+	return this;
+}
```
drbergman, line 1258: do we not want these to do `inherit` or `clone`? I'm speaking from ignorance here, but that seems to be something that we expect to happen every time we copy around `Intracellular` objects

core/PhysiCell_cell.cpp
drbergman: Found where further entries are updated below. I'm wondering now if this here is dead code.

drbergman (commented):
Looks 90% ready! Last 10% for me is just clarifying a couple changes. See individual comments. 

For some of these comments, I'm unclear on the distinctions between `start`, `clone`, and `inherit`. If someone who knows how those work approves, I will approve regarding those comments, too.

core/PhysiCell_cell.cpp
marcorusc: Looks like the code is duplicated?
Also, a small note: for the new dFBA update, I had to create a new instance every time because the inheritance part was causing some issues. So for dFBA it is going to be just:

`#ifdef ADDON_PHYSIDFBA
		if (model_type == "dfba") {
			// Always create a new instance per cell definition
			PhysiCelldFBA::dFBAIntracellular* pIntra = new PhysiCelldFBA::dFBAIntracellular(node);
			pCD->phenotype.intracellular = pIntra->getIntracellularModel();
		}
#endif`

vincent-noel:
Sorry, only seing this now. 
I will start by looking asap at #425 to have a better idea about when to use start().
About inherit, it's goal is really to implement the inheritance of intracellular characteristic from the model cell: 
- in the case of PhysiBoSS, which Boolean variable needs to be reset, and which needs to be copied from the mother ?
- in the case of roadrunner, some quantities (receptor, pool of proteins in cytoplasm) should be halved between the mother and the daughter ?

So to me it's very different from the clone/start issue, and should only be called at division.

core/PhysiCell_cell.cpp
vincent-noel: I was wondering the exact same thing when I rebased it a few days ago. A long time ago, we had this mechanism for inheriting cell definitions, but Paul decided removing it. I am not 100% it is completely removed, but this might be dead code from this old mechanism.

PR #416, opened 2026-05-18
Add support for custom SVG cell colors and validation in PhysiCell settings

drbergman:
### Motivation

`paint_by_number_cell_coloring()` previously had two limitations:

1. **Hard cap at 13 cell types.** Models with 14 or more cell types would render all extra cells as white — invisible against a white background and indistinguishable from each other.
2. **No user control.** Colors were fully hard-coded; users who wanted a specific cell type to appear in a specific color had no mechanism to express that in the config file.

### What changed

**Automatic color generation for extra cell types** (always active)

The 13-color hard cap is removed. For any model with more than 13 cell types and no user-specified colors, extra types are now assigned colors automatically using golden-angle HSV spacing (137.5° hue steps, saturation 0.65, value 0.85). This guarantees perceptual separation between consecutive colors and produces the same sequence regardless of how many cell types the model has.

**Optional per-cell-type color overrides via XML**

A new optional `<cell_colors>` element can be placed inside `<save><SVG>` in the config XML. Each child `<cell_color name="...">` maps a cell definition name to an SVG color string. Partial coverage is supported — only the cell types you want to override need entries; all others fall back to the automatic assignment.

When `<cell_colors>` is present, colors that the user has explicitly claimed are removed from the automatic pool before it is handed to unspecified cell types. For example, if `my_special_cell_type` is assigned `green`, then the remaining cell types cycle through `grey`, `red`, `yellow`, *(green skipped)*, `blue`, etc.

**Validation**

Two error conditions halt the simulation with a clear message:
- A `<cell_color name="...">` whose `name` does not match any cell definition in the model.
- A color value that is not a recognized SVG color (not a CSS named color, `#rrggbb`/`#rgb` hex, or `rgb(r,g,b)` functional notation). This catches typos like `"corral"` before they silently render as black.

**Files changed**

| File | Change |
|------|--------|
| `modules/PhysiCell_settings.h` | Added `svg_cell_colors_specified` flag and `svg_cell_colors_by_name` map to `PhysiCell_Settings` |
| `modules/PhysiCell_settings.cpp` | Parses the optional `<cell_colors>` block from `<save><SVG>` |
| `modules/PhysiCell_pathology.cpp` | Rewrote color-table setup in `paint_by_number_cell_coloring()`; added `hsv_to_svg_color()` and `is_valid_svg_color()` helpers |
| `modules/PhysiCell_pathology.h` | Added `<algorithm>`, `<unordered_set>` includes |
| `config/PhysiCell_settings.xml` | Added commented usage example inside `<SVG>` |

### Usage

Add a `<cell_colors>` block inside `<save><SVG>` in your config XML. You can specify any subset of your cell types — or none at all.

```xml
<SVG>
    <interval units="min">60</interval>
    <enable>true</enable>
    <cell_colors>
        <cell_color name="cancer cell">red</cell_color>
        <cell_color name="immune cell">cyan</cell_color>
        <cell_color name="stroma">rgb(210,180,140)</cell_color>
        <cell_color name="vessel">#8B0000</cell_color>
    </cell_colors>
</SVG>
```

Accepted color formats:
- CSS named colors: `red`, `coral`, `lightskyblue`, … (full list at [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/named-color))
- Hex: `#rgb` or `#rrggbb`
- Functional: `rgb(r,g,b)` with integers 0–255

Cell types not listed receive colors drawn from the standard palette in order, with any user-claimed colors skipped. If the standard palette is exhausted, additional colors are generated automatically.

### DIFF
--- a/config/PhysiCell_settings.xml
+++ b/config/PhysiCell_settings.xml
@@ -35,6 +35,18 @@
         <SVG>
             <interval units="min">60</interval>
             <enable>true</enable>
+            <!-- Optional: override SVG colors for specific cell types.
+                 Names must match cell_definition names exactly; an unrecognized name is an error.
+                 Partial coverage is fine: unlisted cell types are assigned colors automatically
+                 from the built-in palette (with any user-claimed colors skipped), then generated
+                 via golden-angle HSV spacing once the palette is exhausted.
+                 If this element is absent entirely, all colors are assigned automatically.
+                 Accepted color formats: CSS named colors (e.g. "red"), #rrggbb/#rgb hex, rgb(r,g,b).
+            <cell_colors>
+                <cell_color name="cancer cell">red</cell_color>
+                <cell_color name="immune cell">cyan</cell_color>
+            </cell_colors>
+            -->
         </SVG>
         <legacy_data>
             <enable>false</enable>
--- a/modules/PhysiCell_pathology.cpp
+++ b/modules/PhysiCell_pathology.cpp
@@ -1707,40 +1707,197 @@ std::string paint_by_density_percentage(double concentration, double max_conc, d
 	return colormap[ind];
 }
 
+// Returns true if the string is a valid SVG/CSS color:
+//   - CSS named color (case-insensitive)
+//   - #rgb or #rrggbb hex notation
+//   - rgb(r,g,b) functional notation (integers 0-255)
+static bool is_valid_svg_color( const std::string& s )
+{
+	if( s.empty() ) return false;
+
+	// #rgb / #rrggbb
+	if( s[0] == '#' )
+	{
+		if( s.size() != 4 && s.size() != 7 ) return false;
+		for( size_t i = 1; i < s.size(); i++ )
+		{
+			char c = s[i];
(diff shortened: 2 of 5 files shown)


jeanettejohnson:
this will allow people to map colors however they want which is awesome, but does that option only kick in over 13 cell types? This looks great to me

drbergman:
> this will allow people to map colors however they want which is awesome, but does that option only kick in over 13 cell types? This looks great to me

See:
> Optional per-cell-type color overrides via XML
>
> A new optional <cell_colors> element can be placed inside <save><SVG> in the config XML. Each child <cell_color name="..."> maps a cell definition name to an SVG color string. Partial coverage is supported — only the cell types you want to override need entries; all others fall back to the automatic assignment.
> 
> When <cell_colors> is present, colors that the user has explicitly claimed are removed from the automatic pool before it is handed to unspecified cell types. For example, if my_special_cell_type is assigned green, then the remaining cell types cycle through grey, red, yellow, (green skipped), blue, etc.

PR #423, opened 2026-08-06
Fix rules engine treating behavior/signal synonyms as separate entries

heberlr:
## Summary
- Rules that reference the same behavior or signal via different synonyms
  (e.g. `cycle entry` vs `exit from cycle phase 0`) were being stored as two
  separate `Hypothesis_Rule` entries instead of one merged multivariate Hill
  response, because the rules engine keyed its internal maps by the raw
  string from the CSV/XML rule instead of the canonical, synonym-resolved
  name. Whichever rule applied last at runtime silently overwrote the other's
  contribution to the phenotype parameter.
- Canonicalize behavior/signal names at the points they're used as map keys
  or compared by name (`add_behavior`, `find_behavior`, `add_signal`,
  `find_signal`, `set_half_max`, `set_hill_power`, `set_response`).
- Fix a related latent bug where `signals_map[signal] = signals_map.size()`
  used map size as a stand-in for vector index; this diverges (and can
  collide with another signal's index) when the same signal is added twice
  with opposite responses in one rule.
- Remove a dead code path in `add_rule` that reassigned a rule's canonical
  `behavior` field back to the caller's raw synonym string, which would have
  undone the canonicalization above.
- Signals as rule *inputs* were not affected by the overwrite bug (each is
  evaluated independently and summed into the Hill function, not merged into
  a shared per-name structure), but did share the raw-string lookup, which is
  now fixed consistently for both.

## How test it

See the attached example model in the comments below (template project, using
two rules for the same behavior via different synonyms — `pressure decreases
cycle entry` and `substrate increases exit from phase 0`) demonstrating the
bug before this fix and the corrected merged behavior after it.

### DIFF
--- a/core/PhysiCell_rules.cpp
+++ b/core/PhysiCell_rules.cpp
@@ -250,8 +250,9 @@ void Hypothesis_Rule::English_display_HTML( std::ostream& os )
 
 void Hypothesis_Rule::add_signal( std::string signal , double half_max , double hill_power , std::string response )
 {
+	int signal_index = find_signal_index(signal);
     // check: is this a valid signal? (is it in the dictionary?)
-    if( find_signal_index(signal) < 0 )
+    if( signal_index < 0 )
     {
         std::cout << "Error! Attempted to add signal " << signal << " which is not in the dictionary." << std::endl; 
         std::cout << "Either fix your model or add the missing signal to the simulation." << std::endl; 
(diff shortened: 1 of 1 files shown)


heberlr:
Attached is the template model to demonstrate the bug.


[cell_rules.csv](https://github.com/user-attachments/files/30803541/cell_rules.csv)
[cells.csv](https://github.com/user-attachments/files/30803542/cells.csv)
[PhysiCell_settings.xml](https://github.com/user-attachments/files/30803543/PhysiCell_settings.xml)

PR #427, opened 2026-08-07
Thread-reproducible simulation: counter-based RNG + race fixes

heberlr:
## Summary

Adds an opt-in, deterministic RNG mode (Philox4x32-10, keyed by `(cell_id, time_step, purpose, sub_index)`) so a simulation produces bit-identical output regardless of OpenMP thread count — and fixes four core-engine race conditions that RNG determinism alone doesn't address, since they involve one cell's thread mutating another cell's live state, not random draws.

Off by default (`<options><rng_mode>counter_based</rng_mode></options>` to enable); existing projects are unaffected unless they opt in.

## Why

Thread-count-dependent (and in some cases run-to-run-dependent, at a fixed thread count) results meant there was never a reliable way to tell a genuine model bug from scheduling noise, or to reproduce a specific run for debugging. Every race fixed here was a real, silent correctness bug — not just a reproducibility gap — that had gone undetected because nothing before could reliably say "these two runs should match and don't." See `protocols/counter_based_rng.md` for the full design writeup, including why each fix needs `ordered` vs. just `critical`, and why fixes are gated behind `use_counter_based_rng` rather than applied unconditionally.

## What's included

**RNG core** (`modules/philox.h`, `core/PhysiCell_utilities.{h,cpp}`): a keyed `Random(cell_id, time_step, purpose, sub_index)` entry point on top of D. E. Shaw Research's Philox4x32-10 (vendored from [Random123](https://www.deshawresearch.com/resources_random123.html); see Credit section in the doc), plus a thread-local deterministic context so existing `UniformRandom()`/`NormalRandom()`/etc. call sites — core and user code alike — pick it up automatically without being rewritten.

**Simulation-loop wiring** (`modules/PhysiCell_settings.{h,cpp}`, `core/PhysiCell_cell_container.{h,cpp}`): the `<rng_mode>` config option, and a deterministic context installed around each per-cell update phase, keyed by phase-specific purpose. Division/death are applied serially in ID-sorted order (when enabled) so new cell IDs and cell ordering stay thread-count independent — the existing pattern the four fixes below extend.

**Four cross-cell mutation races found and fixed**, discovered via repeated 1-vs-4-thread comparisons on real, multi-behavior projects, not by inspection:
1. `standard_cell_cell_interactions()` (attack/ingest/fuse) — mutates a target cell with no commit ordering; two attackers can race on the same target.
2. `dynamic_spring_attachments()` — checks a neighbor's attachment capacity before `attach_cells_as_spring()` re-validates it; two cells can both attach to an already-full neighbor.
3. `max_cell_interactive_distance_in_voxel` — unprotected read-compare-write on a shared per-voxel bound, in `convert_to_cell_definition()` and (adopting upstream [PR #409](https://github.com/MathCancer/PhysiCell/pull/409)) `update_voxel_in_container()`.
4. `Secretion::advance()` → `simulate_secretion_and_uptake()` — accumulates directly into a shared voxel density vector with no lock; a genuine lost-update on substrate mass, independent of RNG mode, that predates this branch.

Fixes 1, 2, and 4 use `#pragma omp ordered` (the operation isn't commutative — order matters); fix 3 only needs `critical` (max is order-independent). Fixes 1/2/4 are gated behind `use_counter_based_rng`: `ordered` isn't free (DOACROSS-style serialization of the marked region, see literature discussion in the doc), so the legacy path keeps its original, untouched, fully-parallel loop. That means these three races are still present with `counter_based_rng` off — a deliberate, documented trade-off, not an oversight.

**Testing tool** (`beta/test_thread_repro.py`): runs a built project at two thread counts with the same seed and diffs the final output — the tool that surfaced all four races above.

**Docs** (`protocols/counter_based_rng.md`): full design rationale, why each fix needs the synchronization primitive it uses, why gating was chosen over unconditional fixes, how to enable the feature, and how to test a new/modified project for thread reproducibility.


modules/PhysiCell_settings.cpp
```diff
@@ -276,6 +276,27 @@ void PhysiCell_Settings::read_from_pugixml( void )
...
+				PhysiCell_settings.use_counter_based_rng = false;
+				std::cout << "Using legacy RNG mode" << std::endl;
+			}
+			else if( rng_mode != "" )
+			{
+				std::cout << "ERROR: unsupported rng_mode '" << rng_mode << "'. Use 'legacy' or 'counter_based'." << std::endl;
+				exit(-1);
+			}
+		}
+
```
drbergman, line 299: Could we document here what happens if the config file does not have an `rng_mode` node? I assume it is `use_counter_based_rng = false`

drbergman: and I did just see the answer in the header file 😆 . I still think that showing up here would be helpful as this becomes the one place to look for what it gets set to

heberlr: That is correct. By default, XML without `rng_mode` has `use_counter_based_rng = false`.

rheiland:
Finally tested (on Mac) and like it! It definitely runs a sim more slowly, but I think it will mostly be used for debug anyway. I also tested on Windows and it reproduces results there too. However, my initial comparison between the Mac and Windows results showed they were different. We'd want to confirm this and document accordingly.

PR #430, opened 2026-08-14
Fix/heberlr/pr neighbor list for non movable cells

vincent-noel:
New pr for #410

### DIFF
--- a/core/PhysiCell_standard_models.cpp
+++ b/core/PhysiCell_standard_models.cpp
@@ -645,7 +645,14 @@ void standard_update_cell_velocity( Cell* pCell, Phenotype& phenotype, double dt
 			pCell->add_potentials(*neighbor);
 		}
 	}
-
+	
+	// Set the velocity to zero if the cell is not movable - This allows non-movable cells have neighbors list and simple_pressure calculated, but not move.
+	if( pCell->is_movable == false )
+	{ 
+		pCell->velocity = {0.0, 0.0, 0.0}; 
+		return;
+	}
+	
 	pCell->update_motility_vector(dt); 
 	pCell->velocity += phenotype.motility.motility_vector; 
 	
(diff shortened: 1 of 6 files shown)


core/PhysiCell_standard_models.cpp
```diff
@@ -645,7 +645,14 @@ void standard_update_cell_velocity( Cell* pCell, Phenotype& phenotype, double dt
...
 	}
-
+	
+	// Set the velocity to zero if the cell is not movable - This allows non-movable cells have neighbors list and simple_pressure calculated, but not move.
+	if( pCell->is_movable == false )
+	{ 
+		pCell->velocity = {0.0, 0.0, 0.0}; 
+		return;
+	}
+
```
drbergman: This differs from #410. @heberlr made a [change](https://github.com/MathCancer/PhysiCell/pull/410/changes) in core/PhysiCell_cell.cpp.

It looks like the choice here is to let the `axpy( &velocity , temp_r , displacement );` update the `velocity` field. Then in this function, sets that `velocity` to **0** and short circuits the motility updates.

So I think this does the same thing for the most part:
- `add_potentials` will now update the cell's `velocity`. Heber's version won't.
- This version skips the motility vector update and sets `velocity` to **0;** Heber's does the motility vector update. Looking more closely, I think it is technically possible for a cell to have `is_motile==true` but `is_movable==false`. So this particular PR better protects against that.

drbergman (approved):

core/PhysiCell_standard_models.cpp
heberlr: Thanks @vincent-noel! I believe you got the first commits where I used this approach, but later I changed it to do exactly what @drbergman mentioned (see commit 7ffa1a6: https://github.com/MathCancer/PhysiCell/pull/410/changes/7ffa1a65cb02cc22968857c98b62ae91ee4573e), avoiding the extra calculation.

drbergman: OK, so let's put that guard inside `add_potentials` back and then we'll be ready to merge once we've updated the test suite. Sound good @vincent-noel ?

vincent-noel:
Thanks for catching this @drbergman and @heberlr. 
I added the missing commit. All good now, right ?

heberlr (approved):

PR #432, opened 2026-08-16
warn when a transformation hands an attack to a non-attacking type

drbergman:
Draft. **Stacks on #341** — the first two commits here are that PR's; the new work is the last commit. Review/merge #341 first and this will shrink to one commit.

### Why

#341 makes an in-progress attack survive the attacker's transformation. That attack can land on a cell definition that would never have started it, which is the concern raised on #341:

- `attack_rates` default to **0**, so a type the user never wrote attack parameters for inherits the attack anyway
- `attack_damage_rate` defaults to **1.0** and `attack_duration` to **30 min**, so those defaults — not the model — decide how much damage gets done

This is easy to arrive at by accident: it needs no attack configuration on the new type at all.

### What this does

Nothing to the simulation. Whether such an attack should continue is a modelling question, and every automatic answer I could come up with (drop the attack when the new type's attack rate is zero, add a `conserved`-style flag, re-roll the duration) decides something on the user's behalf. So this only warns, and leaves the call to them.

The warning fires when the new type's attack rate against the target's cell type is zero — i.e. the cell is now continuing an attack it could not itself have begun:

```
Warning: a 'CD8+ T cell' transformed into a 'neutrophil' while attacking a 'bacteria',
	but 'neutrophil' has an attack rate of 0 against 'bacteria', i.e. it would never start this attack itself.
	The attack continues regardless, for up to attack_duration = 0.1 min at attack_damage_rate = 1.
	If that is not what you intend, set the attack parameters for 'neutrophil' explicitly, or end the attack in a custom function.
	(reported once per combination of transformation and target type)
```

Reported once per `(old type, new type, target type)` combination — the same transformation recurs constantly in a real model, and per-cell-per-step output would bury everything else.

### Notes for review

- **Thread safety.** `convert_to_cell_definition` runs inside the parallel phenotype loop. The dedupe list is guarded by a *named* critical: `detach_cells_as_spring()` and the `attacked_by` helpers take the unnamed one, and OpenMP criticals are not reentrant.
- The target's type is read as an `int` through `find_cell_definition_index()` rather than reading `pTarget->type_name`, since a target transforming on another thread in the same step would be rewriting that string. `find_cell_definition_index` only calls `find()` on a map that is fixed after setup.
- `attack_rates` is indexed via `find_cell_definition_index`, not by `type` directly — `Cell::type` is the XML ID, which is not necessarily the definition index. (`Cell_Interactions::attack_rate(name)` was not reused here: it does `unordered_map::operator[]`, which inserts on a miss and so is not safe to call concurrently.)

### Verification

Built and ran `interaction-sample` with `CD8+ T cell -> neutrophil` transformation enabled and a 30 min attack duration:

- warning fires **once** across multiple qualifying transformations
- **silent** when `neutrophil`'s attack rate against `bacteria` is set > 0
- clean at `OMP_NUM_THREADS=8`

`interactions` and `rules_sample` are the only sample configs with both nonzero attack and transformation rates. Both run to completion on their stock configs with **zero** warnings, so this does not add noise to existing models.

### Open question

Should this be suppressible from the config, or is once-per-combination quiet enough?

### DIFF
--- a/core/PhysiCell_cell.cpp
+++ b/core/PhysiCell_cell.cpp
@@ -1158,12 +1158,11 @@ void Cell::convert_to_cell_definition( Cell_Definition& cd )
 	Molecular cell_molecular = phenotype.molecular;
 	Custom_Cell_Data cell_custom_data = custom_data;
 
-	// pAttackTarget is phenotype state, and the assignment below replaces it, so
-	// end this cell's own attack. attacked_by and the spring are in state, which
-	// transformation preserves, so attacks against this cell continue.
-	remove_self_from_attacked();
-	// should we also remove all attackers?  That would be a change in behavior, so for now we don't.
-	// remove_all_attackers();
+	// pAttackTarget is phenotype state, and the assignment below replaces it, so hold
+	// on to it and put it back afterwards. attacked_by and the spring are in state,
+	// which transformation preserves, so restoring pAttackTarget keeps the whole
+	// attack link intact -- matching attacks against this cell, which already persist.
+	Cell* cell_attack_target = phenotype.cell_interactions.pAttackTarget;
 
 	// use the cell defaults; 
 	type = cd.type; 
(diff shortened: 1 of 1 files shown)


rheiland:
I wonder if we should adopt the creation/use of a  std::ofstream  warnings.log file? The concern has always been that warnings from cout will simply fly by the user. There will likely be more warnings that could go into such a .log file. At least users could check it post-sim. Unfortunately this feature would affect all sample projects... or else we just show it's possible in the template project.

drbergman:
I think that's a great idea. If we can build it entirely apart from user files, then we don't need this to touch sample projects. We could demonstrate how to hook into it for main.cpp and custom_modules in the template project.

